Issue #122, opened 2015-11-27
clang does not compile re2c 0.15.x


otaran:
### Steps To Reproduce

1. Download https://github.com/skvadrik/re2c/archive/0.15.tar.gz, unpack and `cd` to root directory
2. Run `git init` (otherwise `./re2c/autogen.sh` will fail with error `fatal: Not a git repository (or any of the parent directories): .git`)
3. Run `cd re2c` (otherwise `./re2c/autogen.sh` will fail with error `autoreconf: 'configure.ac' or 'configure.in' is required`)
4. Run `./autogen.sh`
5. Run `./configure`
6. Run `make` and observe

### Expected Result

`make` runs fine and makes `re2c` binary

### Actual Result

`make` fails with following error:

```shell
src/codegen/emit_dfa.cc:250:65: error: use of overloaded operator '<<' is ambiguous (with operand types 're2c::OutputFile' and 'const size_t'
      (aka 'const unsigned long'))
        o << indent(ind++) << "static void *" << opts->yyctable << "[" << conds << "] = {\n";
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~
./src/codegen/output.h:84:22: note: candidate function
        friend OutputFile & operator << (OutputFile & o, char c);
                            ^
./src/codegen/output.h:85:22: note: candidate function
        friend OutputFile & operator << (OutputFile & o, uint32_t n);
                            ^
./src/codegen/output.h:86:22: note: candidate function
        friend OutputFile & operator << (OutputFile & o, uint64_t n);
                            ^
```

### Notes

Reproducible with `re2c` versions `0.15`, `0.15.1` and `0.15.2`

OS: Apple OS X El Capitan (10.11.1)
Developer Tools: Xcode 7.1.1 (7B1005)

```shell
$ clang++ --version
Apple LLVM version 7.0.0 (clang-700.1.76)
Target: x86_64-apple-darwin15.0.0
Thread model: posix
```

skvadrik:
Hi Oleksii!

[...]

PR #123, opened 2015-11-28
Fix building on OS X (see #122)

otaran:


re2c/src/parse/parser.ypp
```diff
@@ -159,7 +159,7 @@ void default_rule(CondList *clist, const Code * code)
 
 %token CLOSE
 %token CLOSESIZE
-%token CODE
+%token TOKEN_CODE
```
otaran: Renamed this to `TOKEN_CODE` because it would produce `#define CODE ...` which in turn would collide with other declarations like `opt_t::CODE` or `OutputFragment::CODE`.

re2c/src/util/u32lim.h
```diff
@@ -23,6 +23,9 @@ class u32lim_t
 	explicit u32lim_t (uint64_t x)
 		: value (x < LIMIT ? static_cast<uint32_t> (x) : LIMIT)
 	{}
+	explicit u32lim_t (size_t x)
```
otaran: Added constructor for `size_t` because on OS X `size_t` is neither `uint32_t` nor `uint64_t`.

skvadrik:
Pulled your fix, thanks! See this commit: https://github.com/skvadrik/re2c/commit/c172f266b4b611cb69bde3b46e4be350819cde73

I had to tune it a bit. The fix in ```re2c/src/util/u32lim.h``` actually fixes OS X builds, but breaks builds on my platform: since ```size_t``` is the same as ```uint64_t```, overloaded constructor is invalid (you cannot make two different overloads for the same type). I'll fix it by dropping overloaded constructor in favour of static constructor-like methods with different names.

skvadrik:
Fixed ```re2c/src/util/u32lim.h``` (overloaded constructor resolution) in this commit: https://github.com/skvadrik/re2c/commit/f65caba520107cb88a7258df07fdf2936557b00

PR #131, opened 2016-01-02
Use bash-specific '[[' builtin

trofi:


release.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-if [ $# -ne 1 ]
+if [[] $# -ne 1 ]]
```
skvadrik: [[]? What you think you're doing Dave?

trofi: Ah, right you are! Will test next time :)

skvadrik: I applied changes proposed in your patch: https://github.com/skvadrik/re2c/commit/b50ab619709e3ce55783b629f785d547cd538dd9

Thank you ;)

PR #159, opened 2016-11-05
Fix #158 Inconsistent forward declaration of struct/class vs definition

### LINKED ISSUE #158: Inconsistent forward declaration of struct/class vs definition
structs State and OutputFile forward declared as class.
classes Range, RegExp and RuleOp forward declared as struct.

These cause warnings in MS C++

pauloscustodio:
structs State and OutputFile forward declared as class.
classes Range, RegExp and RuleOp forward declared as struct.

Fixed forward declarations.

### DIFF
--- a/re2c/src/codegen/bitmap.h
+++ b/re2c/src/codegen/bitmap.h
@@ -10,8 +10,8 @@ namespace re2c
 
 struct Go;
 struct Span;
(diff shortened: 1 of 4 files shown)


skvadrik:
Pulled, thank you!

PR #161, opened 2016-11-05
Fix #160 - Open text files with "wb" causes issues on Windows

### LINKED ISSUE #160: Open text files with "wb" causes issues on Windows
After every run of re2c I get a warning in Visual Studio about "inconsistent line ending" on the generated files. The cause is the "b" in fopen(file_name, "wb") that opens the file in binary mode, and stdio does not write the correct line ending for Windows ("\r\n" instead of "\n" in Linux). In Linux the "b" flag has no effect.

The issue happens in two locations: output.cc:74 and output.cc:349

Can this be changed to fopen(file_name, "w")?

I can submit a PR if the answer is positive.

pauloscustodio:
Text files need to be opened for writing with "w", so that stdio does
the right thing in respect to the correct line endings for the current OS.
("\r\n" in Windows, "\n" in Linux).

### DIFF
--- a/re2c/src/codegen/output.cc
+++ b/re2c/src/codegen/output.cc
@@ -71,7 +71,7 @@ bool OutputFile::open ()
 	}
 	else
 	{
-		file = fopen (file_name, "wb");
+		file = fopen (file_name, "w");
 	}
 	return file != NULL;
 }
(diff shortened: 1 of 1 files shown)


skvadrik:
Pulled, thank you!

PR #169, opened 2017-01-23
correct some minor manpage typos

jcfp:
these showed up in lintian during a maintenance update of the debian re2c package


skvadrik:
Pulled (see https://github.com/skvadrik/re2c/commit/dd5a0077b20d426bfdadf10429426e7dd26a17c9), thank you!

PR #178, opened 2017-04-09
Moves rts used in the manpage to master

pskocik:
and modifies the buildscript to bring them back for the build


skvadrik:
Pulled! I slightly changed your script to use `git archive` instead of `git worktree` (the first one is present in older versions of git).

PR #179, opened 2017-04-09
compose manpage out of rsts from gh-pages-gen

pskocik:
uses the rsts moved from gh-pages-gen to compose the manpage


skvadrik:
It's almost ok, except that a couple of Travis CI builds failed: https://travis-ci.org/skvadrik/re2c/builds/220255998

I believe that failures are caused by include paths in `doc/manpage.rst`: they should be relative to top source directory, that is, `doc/manual/options/options_list.rst` instead of `manual/options/options_list.rst` an so on.

skvadrik:
I'll pull your patch and add a fix on top of it.

skvadrik:
Pulled your patch; fixed paths to be relative to `@top_srcdir@` (see https://github.com/skvadrik/re2c/commit/e5c74f4f0c5f71670ff28058e39793c7da632e9b). Now all checks pass: https://travis-ci.org/skvadrik/re2c.

One last question is: why did you choose to name files `.rst_`, not `.rst`?

pskocik:
Thanks for fixing it. I should've had it run through travis too. 

I named the fragments rst_ to prevent them from being picked up by the build system for rst to html conversion.

PR #188, opened 2017-07-01
Add const to operator() of newver_cmp_t

nightlark:
A certain set of compilers gives a compiler error due to losing "const-volatile qualifiers".


skvadrik:
I pulled the first patch: https://github.com/skvadrik/re2c/commit/fa60d948dfe4ba4105853109b8bd81a2214b7e40

I'm not sure about Cmake.  On one hand, I appreciate the effort, and your code might be reused by others. On the other hand, I doubt my ability to support two different build systems: we already tried to support both MSVC and Autotools, and it ended up in a bunch of unusable ancient MSVC stuff hanging around for years. Pulling in a couple of Cmake files probably won't hurt (as long as they are few and don't get mixed up with the primary build system), but they will quickly fall behind, unless you are ready to maintain them on a regular basis.

As for the third patch, I think it is better to rewrite the code.

nightlark:
Ah oops, actually I meant to only send in the first part of that PR (the const keyword addition). Gotta remember to create a separate branch for PRs next time. Though its good to know your thoughts on the other two things. 👍 

Maybe I'll look at the iterator issue again; getting rid of done.resize to e-b might be the easiest way, but working through the corner cases I wasn't sure about the behavior of std::sort being given an iterator that is technically within the bounds of the vector, compared to one that is pointing at something outside the bounds of the resized vector. I think it *should* be okay as long as std::sort doesn't include what the ending iterator points to, but need to double check that.

For the CMake scripts, yea, multiple build systems are hard to maintain. I'm trying to integrate re2c into build system for another project, and I want it to be easy for people to get it set up without installing and building a bunch of dependencies on their own (particularly annoying for getting started developing on Windows...). As the scripts are right now, I need to add support for some some build options and running tests, and add cmake modules (which should rarely need changes) to help other projects use re2c as part of their build scripts.

Qix-:
I know this is grave-digging, but CMake is an all-in-one solution and most new projects use it nowadays. It makes including nested dependencies a breeze compared to other build systems, and you wouldn't have to distribute MSVC projects alongside the source code (instead letting users generate them from CMake itself).

Further, it supports Ninja, XCode, and a slew of other generators.

I really recommend switching to it. It's not perfect but its adoption is really unparalleled.

Issue #140, opened 2016-04-21
Update manual


skvadrik:
Manual on the website http://re2c.org is should be fixed and updated. This issue is a reminder to do it, as well as a place to pile all relevant notes and ideas. TODO:
- explain rule precedence (longest match, earliest rule)

skvadrik:
Describe control flow in case of multiple re2c blocks as requested in https://github.com/skvadrik/re2c/issues/143.

skvadrik:
Add benchmarks (see #149)

skvadrik:
Explain control flow in conditions (see bug #151).

skvadrik:
Add section (tutorial, guidelines, etc.) about lexer termination and buffer refilling. In particular, highlight the following topics:

- `YYFILL()` vs sentinel termination metod
- what must be updated by `YYFILL`
- how to arrange pointers for `YYFILL`: struct vs standalone pointers (include examples of generated assembly for both cases)
- how tag implementation depends on the presence of `YYFILL`: absolute vs relative tags

skvadrik:
Add lower version reminders to avoid confusion like this: http://stackoverflow.com/questions/39738102/syntax-error-from-default-rule/39766309#39766309

skvadrik:
Add skeleton documentation: : http://skvadrik.github.io/aleph_null/posts/re2c/2015-01-15-input_model_custom.html and http://skvadrik.github.io/aleph_null/posts/re2c/2015-01-13-input_model.html

skvadrik:
Complete Wikipedia article: https://github.com/skvadrik/re2c/issues/168

skvadrik:
Document comment syntax: https://github.com/skvadrik/re2c/issues/180

skvadrik:
Document operator precedence: #183

pmetzger:
Moving my documentation issue here per @skvadrik (and although this is different in `master` I think the version there is mangled):

In the man page:

>       re2c:yyfill:enable = 1;
>              Set this to zero to suppress  generation  of  YYFILL  (n).  When
>              using this be sure to verify that the generated scanner does not
>              read behind input. Allowing this behavior might introduce  sever
>              security issues to you programs.

You probably want this to read:

>              Set this to zero to suppress  generation  of  YYFILL  (n).  When
>              using this**,** be sure to verify that the generated scanner does not
>              read **beyond** the **available** input. Allowing this behavior
>              might introduce **severe** security issues to **your** programs.

[...]

PR #189, opened 2017-07-02
Typo fix and small grammatical change

nightlark:
I noticed that the pages on re2c.org seem to match the gh-pages branch, which appears to be out-dated compared to the gh-pages-gen contents. I'm guessing they only getting sync'ed occasionally?


skvadrik:
Pulled, thank you!

>  I'm guessing they only getting sync'ed occasionally?

Generally I update website and docs after release. There's quite a bit of  important new stuff that needs new examples, man sections an so on, and some of the old documentation must be fixed (see #140 ).

PR #190, opened 2017-07-02
Fix for possible invalidated iterator after resize

nightlark:
Avoids potentially invalid iterator after vector resize. Should double check that the logic for this is still sound. From what I can tell, e acts like done.end() and won't be included in the sorted elements. Since e >= j, resize j-b+1 will cover dropping the same elements as resize e-b,  if e != j. Otherwise, if e == j then either all states after the first partition call need to be dropped with resize e-b, or there are no states to be dropped and the resize will do nothing.


skvadrik:
I pulled your fix and tweaked it a bit: https://github.com/skvadrik/re2c/commit/2d8def82cac9226f26da5a5e8a89b0ea96f934e0
Thanks!

PR #191, opened 2017-07-31
Makefile.am: create target directory before writing into it

rossburton:
In some situations src/parse/ may not exist before a file is copied into the
directory. Ensure that this doesn't happen by creating the directory first.

### DIFF
--- a/re2c/Makefile.am
+++ b/re2c/Makefile.am
@@ -252,6 +252,7 @@ $(BOOTSTRAP_PARSER): $(CUSTOM_PARSER)
 		$(top_srcdir)/$(CUSTOM_PARSER);
 
 .re.cc:
+	$(AM_V_at)$(MKDIR_P) $(dir $@)
 	$(AM_V_GEN) if test -x $(RE2C); \
 	then \
 		$(top_builddir)/$(RE2C) $(RE2CFLAGS) -o $@ $< && \


skvadrik:
Pulled, thank you!
I've just ran into (seemingly) the same error (on a very slow VM running NetBSD-7.0), but I can't reproduce the failure.

rossburton:
On 31 July 2017 at 16:22, skvadrik <notifications@github.com> wrote:

> Pulled, thank you!
> I've just ran into (seemingly) the same error (on a very slow VM running
> NetBSD-7.0), but I can't reproduce the failure.
>
Normally the build works fine as automake dependency files are written
during configure so the directory exists, but if they're not enabled then
you can hit the race by building one of the generated files manually.

Ross

trofi:
For posterity before this change applied to re2c build broke in out-of-source mode.
To reproduce the failure it should be enough to pass `--disable-dependency-tracking`
option to `configure`:

```
$ ../re2c/configure --disable-dependency-tracking
$ make src/conf/parse_opts.o V=1
...
if test -x re2c; \
then \
        ./re2c -b -o src/conf/parse_opts.cc ../re2c/src/conf/parse_opts.re && \
        cp src/conf/parse_opts.cc ../re2c/bootstrap/src/conf/parse_opts.cc; \
else \
        cp ../re2c/bootstrap/src/conf/parse_opts.cc src/conf/parse_opts.cc; \
fi
cp: cannot create regular file 'src/conf/parse_opts.cc': No such file or directory
```

https://www.gnu.org/software/automake/manual/automake.html#Dependencies

Thanks for the fix!

PR #192, opened 2017-08-01
Makefile.am: create target directory before writing into it

rossburton:
In some situations src/parse/ may not exist before a file is copied into the
directory. Ensure that this doesn't happen by creating the directory first.

### DIFF
--- a/re2c/Makefile.am
+++ b/re2c/Makefile.am
@@ -228,6 +228,7 @@ CLEANFILES = \
 	$(DOC)
 
 $(AUTOGEN_PARSER): $(CUSTOM_PARSER)
+	$(AM_V_at)$(MKDIR_P) $(dir $@)
 	$(AM_V_GEN) if test $(BISON) = "no"; \
 	then \
 		cp $(top_srcdir)/$(BOOTSTRAP_PARSER) $@ && \


rossburton:
Well this is embarrassing.

Found the identical race ten lines up.

skvadrik:
Pulled, thanks.
> Well this is embarrassing.

Absolutely, it would help if you provided more details how to trigger build failures though. :)

There is a couple of other custom rules in Makefile.am (mostly for building docs), so I expect that it is possible to trigger even more build failures. I'll try to find them all as @trofi suggested in https://github.com/skvadrik/re2c/pull/191#issuecomment-319214510 with ` --disable-dependency-tracking` option.

skvadrik:
Also, you mention `src/parse/` subdirectory --- but there is no such subdirectory in current source tree (it was moved to `src/ast` some time ago). Is it possible that your build failures are caused by unfortunate update/rebuild cycle?

skvadrik:
One more build failure fixed in https://github.com/skvadrik/re2c/commit/00a0b8b073a940cd587465488c7d4776dd1cc821

skvadrik:
Yet another build failure fixed in https://github.com/skvadrik/re2c/commit/f4734c4ff74eb19b60f4b75e7500144ab3988e32

rossburton:
Ah, the errors were hit using the 0.16 release, and I rebased the patches
to master as the rules were identical.

Ross

skvadrik:
Anyway, it helped to reveal potential build failures with `--disable-dependency-tracking`.
I fixed the remaining custom rules in https://github.com/skvadrik/re2c/commit/af3a69bf7eeb9fa206cdd1a34e2238a8af66ccb6.

skvadrik:
@rossburton  also it turns out that `$(dir $@)`  is not portable: it requires secondary expansion feature of GNU make (breaks on bmake). It is better to use POSIX make macro `$(@D)` (documented here: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html, section "Internal Macros").

Fixed in https://github.com/skvadrik/re2c/commit/cd48d98b34e39fd190b48b35ca371c708af4982b.

Issue #211, opened 2018-08-28
re2c -V throws std::out_of_range


floppym:
As reported on Gentoo Linux.

https://bugs.gentoo.org/664752
```
% re2c --version
re2c 1.1

% re2c -V
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::erase: __pos (which is 4) > this->size() (which is 3)
zsh: abort (core dumped)  re2c -V
```

PR #213, opened 2018-08-28
vernum: move version-string-to-vernum converter to a separate helper

trofi:
No functional change. While at it added tests
to cover past failures:
- "1.1": https://github.com/skvadrik/re2c/issues/211
- "0.14": https://sourceforge.net/p/re2c/bugs/55/


floppym:
When I rewrote the `vernum` function, I wondered about the possibility of a version component with 3 or more digits, like `1.123.0`. That sort of version number should probably be rejected somewhere.

skvadrik:
@floppym I'm going to use re2c to implement `vernum` function (as we do with lexer, option parser and configuration parser). This will allow some flexibility and better error handling.

skvadrik:
@trofi  Pulled, thank you! https://github.com/skvadrik/re2c/commit/c17fa5aa485761d8648f65ac95d4a1db47251162

It's good to have a test.

Issue #215, opened 2018-09-03
A memory read overrun issue in s_to_n32_unsafe.cc


mlite:
This is the runtime error msg caused by the overrun. 
```
DTS_MSG: Stensal C/C++ DTS detected a fatal program error!
DTS_MSG: Continuing the execution will cause unexpected behaviors, abort!
DTS_MSG: Reading 1 bytes at 0xffffc7dc will read undefined values.
DTS_MSG: Diagnostic information:

- The object to-be-read (start:0xffffc6dc, size:256 bytes) is allocated at
-     file:/home/sbuilder/workspace/re2c/re2c/src/test/s_to_n32_unsafe/test.cc::39, 10
-  0xffffc6dc               0xffffc7db
-  +------------------------+
-  | the object  to-be-read |......
-  +------------------------+
-                            ^~~~~~~~~~
-        the read starts at 0xffffc7dc that is right after the object end.
- Stack trace (most recent call first):
-[1]  file:/home/sbuilder/workspace/re2c/re2c/src/util/s_to_n32_unsafe.cc::28, 9
-[2]  file:/home/sbuilder/workspace/re2c/re2c/src/test/s_to_n32_unsafe/test.cc::50, 9
-[3]  file:/home/sbuilder/workspace/re2c/re2c/src/test/s_to_n32_unsafe/test.cc::85, 15
-[4]  file:/home/sbuilder/workspace/re2c/re2c/src/test/s_to_n32_unsafe/test.cc::101, 12
-[5]  file:/musl-1.1.10/src/env/__libc_start_main.c::168, 11
```

skvadrik:
Do you have the .re file which caused the failure?

trofi:
Looks like it's caused by the test itself: `re2c/src/test/s_to_n32_unsafe/test.cc`. I haven't found plausible call path yet but a few `use-of-uninitialized-value` reports are detectable by clang's `fsanitize=memory`:

```
$ ./configure CC=clang CXX=clang++ CFLAGS="-fsanitize=memory" CXXFLAGS="-fsanitize=memory"
...
$ make check VERBOSE=1
...
FAIL: run_tests.sh
SKIP: testrange
PASS: testston32unsafe
PASS: testvertovernum
==================================
   re2c 1.1.1: ./test-suite.log
==================================

# TOTAL: 4
# PASS:  2
# SKIP:  1
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

[...]

PR #217, opened 2018-09-04
__alltest.sh: add clang's -fsanitize=memory flavour

trofi:
Bug: https://github.com/skvadrik/re2c/issues/215


skvadrik:
Pulled, thanks. :)

Issue #221, opened 2018-10-13
big source tarball


rofl0r:
hi, i've just updated re2c from 0.14 to 1.1.1 >>[0]<< and noticed the tarball grew from 2.5 to 5.9MB, which is quite big considering the built executable is only 300 KB!
maybe there's some binary documentation inside that bloats the source unnecessarily and could be offered as a secondary download to interested people ?
since folks like me who build their stuff from source usually keep the tarballs around, every MB counts...thanks for considering to shrinking it a bit.
maybe offering a download as tar.xz could also bring the size down a bit.

[0]: https://github.com/sabotage-linux/sabotage/commit/cf7c1ceebd78c11892e933b77807a64f122b6a7a

trofi:
We can check what is actually used in tarball by re-compressing individual directories to get the idea:

```
re2c-1.1.1 $ ls -lh *.gz --sort=size
3.6M test.tar.gz
889K doc.tar.gz
115K src.tar.gz
59K bootstrap.tar.gz
42K examples.tar.gz
```

As we can see these are mostly tests. `6MB` does not sounds like an unreasonable number.

skvadrik:
My conclusion is the same: tests eat all the space (and they will only grow over time). Without tests the resulting tarball is about 100K. We can potentially package stripped tarball without tests, but that requires fiddling with automake conditionals to make things like `make check` pass for the stripped tarball. Is it worth the effort?

@rofl0r: Since you build from source, it is likely that you also want to run the tests, right?

@trofi thanks for measuring!

skvadrik:
I measured `.tar.xz` size, and it is about 2.7M, so maybe using xz the best option.

leo-yuriev:
git-submodule with test's code and data?

rofl0r:
> @rofl0r: Since you build from source, it is likely that you also want to run the tests, right?

not really. all our packages do --disable-tests, when possible, to save time, and to allow cross-compilation (tests work only for native compiles anyway).

skvadrik:
@trofi pushed a fix that switches to .tar.xz by default and reduces tarball size 2x.

> another aspect of having a big source tarball is that it suggests a bloated program inside.
> for example i was evaluating lexers and when re2c was mentioned i looked into my pkg recipe and was immediately repelled by the 2.5 MB tarball size, to the point of immediately abandoning it as a potential candidate. but since i already had it installed for php, i looked a bit closer and figured the program itself is reasonably small.

[...]

PR #222, opened 2018-10-16
configure.ac: enable xz tarballs instead of gzip by default

trofi:
`xz` compresses twice as good as `gzip` on `re2c` sources:

```
$ ls -lh *1.1.1*
4,8M re2c-1.1.1.tar.gz
2,5M re2c-1.1.1.tar.xz
```

Switch `make dist` to `xz by default. `gzip` is still available
via `make dist-gzip`.

Reported-by: rofl0r
Bug: https://github.com/skvadrik/re2c/issues/221


skvadrik:
Thank you @trofi , a very elegant fix! Merged.

Issue #239, opened 2019-01-13
Push model example has wrong re2c invocation, breaks guide


mwberry:
I got frustrated with flex and went looking for a better tool. I found re2c and was glad to see it offered everything flex does and appears to address all the short comings of flex's design. I worked through the examples in order to try to quickly onboard with the re2c syntax and concepts. When I got to the [push model example](http://re2c.org/examples/example_18.html) I found it did not compile, complaining about missing declarations for `YYPEEK`, `YYSKIP`, etc. 

Debugging it was certainly a good exercise and it helped me understand better what state re2c needs to keep track of, but the problem was simply that the wrong arguments are used when invoking re2c. After defining the missing macros, I was surprised to see that `YYGETSTATE` wasn't being used in the generated output. This is because the example invocation uses `--input custom`, hence the missing macros, instead of `--storable-state`, hence the lack of `YYGETSTATE` invocations.

Additionally, there's a typo in the link to the `--storage-state` documentation. It links to [this URL](http://re2c.org/manual/featires/state/state.html) instead of [this url](http://re2c.org/manual/features/state/state.html), with the typo being `s/featires/features/`.

skvadrik:
@mwberry all fixed, thank you for reporting!

PR #240, opened 2019-01-13
[#239] Uses --storable-state argument in push model example

mwberry:
Fixes issue #239. The mistaken '--input custom' argument was replaced
with --storable-state. Additionally a URL typo was fixed


skvadrik:
But your fix was perfect anyway. :D

PR #256, opened 2019-08-13
Package LICENSE file in the release tarball

certik:
This is important when re2c is packaged in distributions, so that one can
quickly check what license the re2c is distributed under. Distributions
typically use the release tarball, not the git repository, and so this commit
puts the LICENSE file into the release tarball also.


skvadrik:
Pulled, thank you!

PR #258, opened 2019-08-13
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds effort](https://reproducible-builds.org/) we noticed that `re2c` could not be built reproducibly. This is because it used the current build date in the manual page and was originally filed in Debian as [#934697](https://bugs.debian.org/934697).

This patch uses the [SOURCE_DATE_EPOCH environment variable](https://reproducible-builds.org/specs/source-date-epoch/).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,8 +6,12 @@ AM_SILENT_RULES([yes])
 AC_CONFIG_SRCDIR([src/main.cc])
 AC_CONFIG_HEADERS([config.h])
 
-
-AC_SUBST(PACKAGE_DATE, `date +'%d %b %Y'`)
+if test -n "$SOURCE_DATE_EPOCH"; then
+	PACKAGE_DATE=`LC_ALL=C date --utc --date="@$SOURCE_DATE_EPOCH" +'%b %d %Y'`
+else
+	PACKAGE_DATE=`date +'%d %b %Y'`
+fi
+AC_SUBST(PACKAGE_DATE, $PACKAGE_DATE)
 AC_SUBST(PACKAGE_VERSION)
 AC_SUBST(PACKAGE_NAME)
 AC_SUBST(PACKAGE_TARNAME)


skvadrik:
@lamby Merged, thank you.

However, I need to make it compatible with [POSIX date](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/date.html), as re2c is built on other platforms as well. Can you suggest a fix? From the [guidelines](https://reproducible-builds.org/docs/source-date-epoch/):

> If you need to support BSD date as well you should fallback to trying ther -r seconds timestamp variant:
```
DATE_FMT="%Y-%m-%d"
SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH:-$(date +%s)}"
BUILD_DATE=$(date -u -d "@$SOURCE_DATE_EPOCH" "+$DATE_FMT" 2>/dev/null || date -u -r "$SOURCE_DATE_EPOCH" "+$DATE_FMT" 2>/dev/null || date -u "+$DATE_FMT")
```
This is not completely POSIX-compliant as well (because of the `-d` and `-r` options).

lamby:
IIRC whilst you are entirely correct about the lack of `d` and `-r` is POSIX there is, unfortunately, no way around this. POSIX is minimal after all. I suspect one of the `||`/OR'd will capture 99.99999% of platforms. I guess if you are still worried then you could do something like an additional suffix of `|| echo "[some kind of fallback]"`

skvadrik:
@lamby Thanks, I think you are right about 99.99999%. However, I realized that `PACKAGE_DATE` was used only in one place (in the manpage) and it was not really needed there. So it was easier to get rid of it altogether: https://github.com/skvadrik/re2c/commit/7648f8dd9eaee660ca85a003e7246ede7b987e63 (thanks to @trofi  for suggesting the idea).

PR #265, opened 2019-12-27
.travis.yml: enable ccache to speed up incremental builds

trofi:
I'm not sure what would be the actual benefit. Let's try it in action!


skvadrik:
Thank you!

PR #277, opened 2020-04-05
Add build scripts using CMake

ligfx:
Follow-up to #275.

This adds `__cmakebuild*.sh` scripts analogous to the existing `__build*.sh` scripts, for a variety of scenarios.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,8 +92,11 @@ ac_config_headers("config.h")
 
 # Makefile.am
 set(RE2C_STDLIB_DIR "${CMAKE_INSTALL_PREFIX}/share/re2c/stdlib")
-add_compile_definitions("RE2C_STDLIB_DIR=\"${RE2C_STDLIB_DIR}\"")
-add_compile_definitions($<$<CONFIG:Debug>:RE2C_DEBUG>)
+add_compile_definitions(
+  "RE2C_STDLIB_DIR=\"${RE2C_STDLIB_DIR}\""
+  $<$<CONFIG:Debug>:RE2C_DEBUG>
+  $<$<CONFIG:RelWithDebInfo>:RE2C_DEBUG>
+)
 include_directories(. "${CMAKE_CURRENT_BINARY_DIR}")
 
 # sources
--- a/__cmakebuild.sh
+++ b/__cmakebuild.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+builddir=__build
+rm -rf $builddir
+mkdir $builddir
+
+cd $builddir
+cmake .. \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+    -DRE2C_BUILD_LIBS=yes \
+    && cmake --build . -j$(nproc)
+cd ..
--- a/__cmakebuild_asan.sh
+++ b/__cmakebuild_asan.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+builddir=__build_asan
+rm -rf $builddir
+mkdir $builddir
+
+cd $builddir
+cmake .. \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+    -DRE2C_BUILD_LIBS=yes \
+    -DCMAKE_CXX_FLAGS="-fsanitize=address" \
+    -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address" \
+    -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address" \
+    && cmake --build . -j$(nproc)
+cd ..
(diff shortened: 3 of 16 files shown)


skvadrik:
Can you please fix two things:

- Add `-DCMAKE_CXX_FLAGS="-O2"` to all scripts except the one that builds in release mode (most of them already set `CMAKE_CXX_FLAGS`, so just add `-O2` there). This will correspond to the behavior of existing scripts, which have `-O2` enabled.

- For Mingw builds, it is necessary to set a few more things (like paths to include files), otherwise Mingw uses wrong headers and the build fails (on my Linux system at least). Instead of `-DCMAKE_SYSTEM_NAME=Windows -DCMAKE_C_COMPILER=i686-w64-mingw32-gcc -DCMAKE_CXX_COMPILER=i686-w64-mingw32-g++`  use `-DCMAKE_TOOLCHAIN_FILE=cmake/Toolchain-cross-mingw32-linux.cmake` and add the following file `cmake/Toolchain-cross-mingw32-linux.cmake`:
```cmake
# the name of the target operating system
SET(CMAKE_SYSTEM_NAME Windows)

SET(COMPILER_PREFIX "i686-w64-mingw32")

# which compilers to use for C and C++
find_program(CMAKE_RC_COMPILER NAMES ${COMPILER_PREFIX}-windres)
find_program(CMAKE_C_COMPILER NAMES ${COMPILER_PREFIX}-gcc)
find_program(CMAKE_CXX_COMPILER NAMES ${COMPILER_PREFIX}-g++)

# here is the target environment located
SET(CMAKE_FIND_ROOT_PATH ${CMAKE_FIND_ROOT_PATH} /usr/${COMPILER_PREFIX})

# adjust the default behaviour of the FIND_XXX() commands:
# search headers and libraries in the target environment, search.
# programs in the host environment
SET(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
SET(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
```
And also a similar file `cmake/Toolchain-cross-mingw64-linux.cmake` for Mingw64:
```cmake
# the name of the target operating system
SET(CMAKE_SYSTEM_NAME Windows)

SET(COMPILER_PREFIX "x86_64-w64-mingw32")

# which compilers to use for C and C++
find_program(CMAKE_RC_COMPILER NAMES ${COMPILER_PREFIX}-windres)
find_program(CMAKE_C_COMPILER NAMES ${COMPILER_PREFIX}-gcc)
find_program(CMAKE_CXX_COMPILER NAMES ${COMPILER_PREFIX}-g++)

# here is the target environment located
SET(CMAKE_FIND_ROOT_PATH ${CMAKE_FIND_ROOT_PATH} /usr/${COMPILER_PREFIX})

# adjust the default behaviour of the FIND_XXX() commands:
# search headers and libraries in the target environment, search.
# programs in the host environment
SET(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
SET(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
```

ligfx:
- > Add -DCMAKE_CXX_FLAGS="-O2" to all scripts except the one that builds in release mode (most of them already set CMAKE_CXX_FLAGS, so just add -O2 there). This will correspond to the behavior of existing scripts, which have -O2 enabled.
   
   What do you think about using `-DCMAKE_BUILD_TYPE=RelWithDebInfo` instead? That expresses the semantics of "an optimized build with debug symbols."

- ~~For Mingw builds, it is necessary to set a few more things~~ Added toolchain files and fixed the __cmakebuild_mingw.sh script

skvadrik:
> What do you think about using `-DCMAKE_BUILD_TYPE=RelWithDebInfo` instead? That expresses the semantics of "an optimized build with debug symbols."

That would be ok if you add `add_compile_definitions($<$<CONFIG:RelWithDebInfo>:RE2C_DEBUG>)`. The main purpose of the build scripts is to provide a fast way of building in a "development configuration", which means 1) with `RE2C_DEBUG`, 2) with `-g` and 3) with `-O2`.

ligfx:
- ~~That would be ok if you add `add_compile_definitions($<$<CONFIG:RelWithDebInfo>:RE2C_DEBUG>)`. The main purpose of the build scripts is to provide a fast way of building in a "development configuration", which means 1) with `RE2C_DEBUG`, 2) with `-g` and 3) with `-O2`.~~ Okay, I added the `RE2C_DEBUG` flag in `RelWithDebInfo` builds, and changed all the scripts except `__cmakebuild_nodebug.sh` to use it. The `RelWithDebInfo` build type automatically adds `-g` and `-O2`, or the equivalent options depending on the platform.

skvadrik:
Merged: https://github.com/skvadrik/re2c/commit/80cd01b58751e7f0ae339ee74d9ff96468f69fae, thank you! (Github doesn't show this pull-request as merge probably because I cherry-picked to keep history linear.)

Issue #284, opened 2020-05-30
mismatched YYBACKUP and YYRESTORE


fanf2:
Hello! I'm trying to use re2c to improve the lexer in unifdef. I
really like the way re2c works, but I have run into a problem, and I'm
not sure if I have made a mistake or misunderstood something or if it
is a bug in re2c.

When I run the program below, the lexer generated by re2c calls
`YYRESTORE()` without previously calling `YYBACKUP()`. I thought that
this should not happen, because all the examples in the documentation
start the lexer with `YYMARKER` uninitialized. I cautiously initialize
`YYMARKER` to NULL, which means that after a mismatched `YYRESTORE()`
the lexer tries to dereference a NULL pointer.

The same thing happens with both re2c 1.2 and 1.3.

I have tried this program with and without `--input custom` and it
seems to behave the same in both cases. However, I'm not confident
that it is compiled correcctly when it is built without `--input
custom` because in that case I can't protect against undefined
behaviour from dereferencing NULL.

I tried to initialize `YYMARKER` to the same as `YYCURSOR` instead of NULL,
and that causes the complete version of my lexer to go into a loop.

I found this problem by testing my rather large lexer with LLVM
libfuzzer. I have tried to find something close to a minimal example
that reproduces the failure mode.

Here is the source code (called `2020-05-30.re`):

```
#include <stdio.h>

#define YYCTYPE		char
#define YYPEEK()	(YYCURSOR == NULL ? 0 : *YYCURSOR)
#define YYSKIP()	++YYCURSOR
#define YYLESSTHAN(n)	(YYDEBUG(9999,0), YYLIMIT - YYCURSOR < n)
#define YYBACKUP()	(YYDEBUG(8888,0), YYMARKER = YYCURSOR)
#define YYRESTORE()	(YYDEBUG(7777,0), YYCURSOR = YYMARKER)

#define myreturn()	return(YYDEBUG(6666,0), YYCURSOR == 0)

#define V const void *
#define YYDEBUG(st, ch)							\
	printf("re2c base %p cursor %p limit %p "			\
	       "accept %d state %d char %c \n",				\
	       (V)ptr, (V)YYCURSOR, (V)YYLIMIT,				\
	       yyaccept, st, ch)

#define yyaccept 0

int main(void) {
	const char *ptr = "/*";
	const char *YYCURSOR = ptr;
	const char *YYLIMIT  = ptr + 2;
	const char *YYMARKER = NULL;

[...]

PR #285, opened 2020-05-31
Add a C preprocessor lexer to the tests.

fanf2:
As discussed in #284, here is the lexer I was working on when I encountered the YYBACKUP/YYRESTORE bug.

Signed-off-by: Tony Finch <dot@dotat.at>

### DIFF
--- a/test/real_world/cpp.re
+++ b/test/real_world/cpp.re
@@ -1,0 +1,353 @@
+/*
+ * This lexer was written for unifdef, which is a kind of C/C++
+ * preprocessor. It supports some recent C and C++ features.
+ *
+ * It lexes comments and whitespace as separate tokens, partly for the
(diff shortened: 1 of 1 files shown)


skvadrik:
Thank you! This is very helpful. I have tweaked your commit a tiny bit and merged it as https://github.com/skvadrik/re2c/commit/e9764de355fb34fd2e2cda341bab49f925075a12. The only changes I made are:

- Add a test result file so that re2c tests don't fail (including Travis CI that runs on every push).
- Rename the test so that re2c runs it with `-i` option (line directives in large real tests make it difficult to separate real changes from trivial ones when reviewing diffs).

As Github cannot track the relation between my commit and this pull request, I'm closing it manually (it won't show as merged, but in fact it is).

PR #287, opened 2020-06-07
configure.ac: store auxiliary configure files in build-aux/

trofi:
This makes root directory less cluttered. Autoconf-specific
scripts like 'config.guess'. 'install-sh' and friends
are moved to build-aux/ now.


skvadrik:
\o/ Thank you.

Issue #198, opened 2017-11-07
no match for ‘operator!=’ in ‘i != std::vector<_Tp, _Alloc>::rend() [with _Tp = re2c::bitmap_t, _Alloc = std::allocator<re2c::bitmap_t>]()’


ryandesign:
Hello, I'm a developer with the MacPorts package management system, and I noticed that re2c 1.0.2 does not build on Mac OS X 10.5.8 with its gcc version 4.2.1 (Apple Inc. build 5577). The error is:

```
depbase=`echo src/code/bitmap.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
	/usr/bin/g++-4.2 -DHAVE_CONFIG_H -I.   -I/opt/local/include -std=c++98  -W  -Wall  -Wextra  -Weffc++  -pedantic  -Wformat=2  -Wredundant-decls  -Wconversion  -O2  -pipe -Os -arch ppc -MT src/code/bitmap.o -MD -MP -MF $depbase.Tpo -c -o src/code/bitmap.o src/code/bitmap.cc &&\
	mv -f $depbase.Tpo $depbase.Po
In file included from ./src/conf/warn.h:9,
                 from ./src/conf/opt.h:9,
                 from ./src/code/output.h:14,
                 from ./src/code/go.h:8,
                 from src/code/bitmap.cc:5:
./src/util/free_list.h: In instantiation of ‘re2c::free_list<re2c::Code*>’:
./src/re/rule.h:32:   instantiated from here
./src/util/free_list.h:11: warning: base class ‘class std::set<re2c::Code*, std::less<re2c::Code*>, std::allocator<re2c::Code*> >’ has a non-virtual destructor
src/code/bitmap.cc: In member function ‘void re2c::bitmaps_t::insert(const re2c::Go*, const re2c::State*)’:
src/code/bitmap.cc:30: error: no match for ‘operator!=’ in ‘i != std::vector<_Tp, _Alloc>::rend() [with _Tp = re2c::bitmap_t, _Alloc = std::allocator<re2c::bitmap_t>]()’
./src/re/tag.h: At global scope:
./src/re/tag.h:15: warning: ‘re2c::TAGVER_BOTTOM’ defined but not used
./src/re/tag.h:17: warning: ‘re2c::TAGVER_CURSOR’ defined but not used
make[1]: *** [src/code/bitmap.o] Error 1
```

I am not a C++ programmer so I can't offer any suggestions.

[...]

PR #292, opened 2020-07-04
Fix for old compilers like gcc 4.2 and clang 3.4

ryandesign:
Hi, I'm the MacPorts maintainer of re2c. We discussed a problem a few years ago in #198 where old versions of gcc were unable to compile your code, and you committed a simple fix but warned that it might break in the future.

It did, and we received a couple reports about it. It was reported last year that [re2c 1.2.1 didn't build with gcc 4.2](https://trac.macports.org/ticket/58993) on Mac OS X 10.4 on PowerPC:

```
./src/dfa/closure_posix.h:128: error: no match for 'operator!=' in 'c != std::vector<_Tp, _Alloc>::rend() [with _Tp = re2c::clos_t, _Alloc = std::allocator<re2c::clos_t>]()'
```

I had forgotten about our previous discussion and worked around the problem by using a newer compiler, but then it was reported a few days ago that [re2c 1.3 didn't build with clang 3.4](https://trac.macports.org/ticket/60755) on Mac OS X 10.4 on Intel:

```
./src/dfa/closure_posix.h:128:61: error: invalid operands to binary expression ('typename determ_context_t<phistory_t>::rcconfiter_t' (aka 'reverse_iterator<const_iterator>') and 'reverse_iterator' (aka 'reverse_iterator<iterator>'))
    for (typename ctx_t::rcconfiter_t c = reach.rbegin(); c != reach.rend(); ++c) {
```

This prompted me to investigate and find our previous discussion again and to try the patch in this PR, which worked for me. I hope you'll accept it.

### DIFF
--- a/src/dfa/closure_posix.h
+++ b/src/dfa/closure_posix.h
@@ -125,7 +125,8 @@ void init_gor1(ctx_t &ctx)
     // init: push configurations ordered by POSIX precedence (highest on top)
     state.clear();
     std::sort(reach.begin(), reach.end(), cmp_gor1_t<ctx_t>(ctx));
-    for (typename ctx_t::rcconfiter_t c = reach.rbegin(); c != reach.rend(); ++c) {
+    typename ctx_t::rcconfiter_t c = reach.rbegin(), e = reach.rend();
+    for (; c != e; ++c) {
         nfa_state_t *q = c->state;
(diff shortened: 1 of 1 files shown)


skvadrik:
Thank you for the fix, and for keeping re2c portable on old compilers!

PR #293, opened 2020-07-04
.travis.yml: split build commands into individual entries

trofi:
This should allow travis CI measure individual times on each step


skvadrik:
Thank you! \o/

PR #304, opened 2020-09-04
Fix CMake command from cmake to ${CMAKE_COMMAND}

s-yasu:
Fix build in environment where CMake is not in path.


skvadrik:
Thanks for the fix!

PR #309, opened 2020-09-28
Initial support of CI using GitHub Actions

sergeyklay:
Hello,

This PR introduces initial GitHub Actions support by adding Configure, Build and Test steps. Right now it uses only Linux. I'll add macOS and Windows support a bit later.

To see how it works now refer to: https://github.com/sergeyklay/re2c/runs/1175055832?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,24 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build-and-test:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Configure
+        run: |
+          cmake -S . -Bcmake-build
+
+      - name: Build
+        run: cmake --build cmake-build
+
+      - name: Test
+        run: cmake --build cmake-build --target check


skvadrik:
Thanks for the PR @sergeyklay! I have a few questions.

As you know, re2c uses Travis CI with an elaborate testing matrix (see [travis.yml](https://github.com/skvadrik/re2c/blob/master/.travis.yml) script and [the latest set of builds](https://travis-ci.org/github/skvadrik/re2c)). It works well and has captured numerous bugs. The build matrix is a killer feature: it allows CMake coexist with Autotools, it runs various sanitizer builds and so on. Do you intend GitHub Actions to replace Travis CI (then would it be possible to migrate travis.yml), or as a complimentary workflow (then what are the additional features it provides that Travis CI doesn't)?

sergeyklay:
Well, yes. It would be nice to have CI alongside of source code. I mean use the same vendor to store source code as well as run CI pipelines.

In the future, I would like to get all-in-one workflow, which covers all the features used in Travis CI as well as:
- macOS builds / tests
- Windows builds / tests (Mingw, Visual Studio)
- Create nightly (unstable) builds
- Create releases and attach artifacts
- Code coverage repoort

Right now it is just a (working) skeleton which is simple enough to add functionality step by step.

skvadrik:
That would be great! Let's have GitHub Actions alongside with Travis for now.

PR #311, opened 2020-10-02
Add Windows support for GitHub Actions

sergeyklay:
Hello,

This PR adds Windows  with Visual Studio 16 2019 to the build matrix on GitHub Action. There still a lot of work, e.g. provide ability to run tests, compiler flags definition and so on. But right now it works. I'll try to improve Windows builds a bit later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,20 +28,15 @@ option(RE2C_BUILD_LIBS "Build libraries" OFF)
 option(RE2C_BUILD_RE2GO "Build re2go executable (an alias for `re2c --lang go`)" ON)
 
 # checks for programs
-if(UNIX)
-    find_program(
-      BISON_EXECUTABLE
-      NAMES bison
-      PATHS /usr /usr/local
-      PATH_SUFFIXES bin)
-elseif(WIN32)
+if(WIN32)
     find_program(
       BISON_EXECUTABLE
       NAMES bison.exe win_bison.exe
       PATHS C:/
       PATH_SUFFIXES "")
+else()
+    find_package(BISON)
 endif()
-mark_as_advanced(BISON_EXECUTABLE)
 
 # checks for C++ compiler flags
 set(CMAKE_CXX_STANDARD 98)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,20 @@ option(RE2C_BUILD_LIBS "Build libraries" OFF)
 option(RE2C_BUILD_RE2GO "Build re2go executable (an alias for `re2c --lang go`)" ON)
 
 # checks for programs
-find_package(BISON)
+if(UNIX)
+    find_program(
+      BISON_EXECUTABLE
+      NAMES bison
+      PATHS /usr /usr/local
+      PATH_SUFFIXES bin)
+elseif(WIN32)
+    find_program(
+      BISON_EXECUTABLE
+      NAMES bison.exe win_bison.exe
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -28,7 +28,20 @@ option(RE2C_BUILD_LIBS "Build libraries" OFF)
...
+      PATHS /usr /usr/local
+      PATH_SUFFIXES bin)
+elseif(WIN32)
+    find_program(
+      BISON_EXECUTABLE
+      NAMES bison.exe win_bison.exe
+      PATHS C:/
+      PATH_SUFFIXES "")
+endif()
+mark_as_advanced(BISON_EXECUTABLE)
```
skvadrik, line 44: Why is this needed? From [the docs](https://cmake.org/cmake/help/latest/command/mark_as_advanced.html?highlight=s):

> An advanced variable will not be displayed in any of the cmake GUIs unless the show advanced option is on. In script mode, the advanced/non-advanced state has no effect.

So I think it affects only how the variable is displayed, but not the build process. In that case let's drop this (it's fine to have it displayed, and it is more important to keep the CMakeLists.txt as simple as possible).

CMakeLists.txt
```diff
@@ -28,7 +28,20 @@ option(RE2C_BUILD_LIBS "Build libraries" OFF)
...
+      NAMES bison
+      PATHS /usr /usr/local
+      PATH_SUFFIXES bin)
+elseif(WIN32)
+    find_program(
+      BISON_EXECUTABLE
+      NAMES bison.exe win_bison.exe
+      PATHS C:/
+      PATH_SUFFIXES "")
+endif()
```
skvadrik, line 43: Can you change this to:
```cmake
if(WIN32)
    find_program(BISON_EXECUTABLE
        NAMES bison.exe win_bison.exe
        PATHS C:/
        PATH_SUFFIXES "")
else()
    find_package(BISON)
endif()
```
Ideally I'd hope that [FindBISON](https://cmake.org/cmake/help/latest/module/FindBISON.html) would work for Windows, but maybe the paths are specific to this setup.

Also, note that bison is only needed to rebuild parser, it shouldn't be needed to build first time (pre-generated bootstrap/src/parse/parser.cc is used for that).

skvadrik (changes requested):
Thanks! I have a few comments for CMakeList.txt changes related to bison.

skvadrik:
By the way, it's the first time I use GitHub review tool, it seems a recent addition. Not bad!

CMakeLists.txt
skvadrik: I tried to add the comment to line 44, but for some reason it shows lines 41-43 as well.

CMakeLists.txt
sergeyklay: I tried various find_package configurations on Windows - no luck :/ Also I noticed all we use below is BISON_EXECUTABLE. 
We just not use the power of find_package. Thus, I replaced find_package by find_program and set BISON_EXECUTABLE explicitly. However, your change looks reasonable and works well. Thank you.

skvadrik: It is true that we do not need the full power of `find_package`. I just thought that `find_package` is the default CMake way and it is expected to encapsulate platform-specific details, but maybe I misunderstand its purpose.

skvadrik (approved):

skvadrik:
Merged, thank you!

sergeyklay:
You're welcome :)

PR #312, opened 2020-10-02
Cleaned up CMakeLists.txt to use platform independent bison search

sergeyklay:
This PR fixes my previous PR (#311) to use platform independent bison search.

After discussion in https://github.com/skvadrik/re2c/pull/311#discussion_r498687041, I decided to double-check this approach again and made sure that it actually works.  I don't quite understand why I wasn't able to get this to work before, maybe I didn't read the documentation carefully or missed something important. Anyway now it works, and looks clear and simple as initially.

To see how it works refer https://github.com/sergeyklay/re2c/runs/1198167217?check_suite_focus=true#step:8:16


skvadrik:
Awesome, thanks for double checking!

PR #315, opened 2020-10-05
Add Release build type to the build matrix

sergeyklay:
Hello,

This PR amends build matrix by adding Release configuration. There are no compiler flags changes yet because I still trying to choice a better way to set flags for various compilers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ defaults:
 jobs:
   build-and-test:
 
-    name: ${{ matrix.toolchain }}
+    name: "${{ matrix.build-type }}: ${{ matrix.toolchain }}"
     runs-on: ${{ matrix.os }}
 
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -70,27 +71,27 @@ jobs:
...
             *)
               cmake -S . -B cmake-build \
-                -DCMAKE_BUILD_TYPE=${{ matrix.configuration }}
+                -DCMAKE_BUILD_TYPE=${{ matrix.build-type }}
               ;;
           esac
 
       - name: Build
-        run: cmake --build cmake-build
+        run: cmake --build cmake-build --config ${{ matrix.build-type }}
```
skvadrik, line 90: Is this needed because both release and debug builds are in the same directory? Then perhaps it is better to pass a different build directory like `cmake -S . -B cmake-build-{{ matrix.build-type }}` to avoid weird bugs due to muti-configuration builds?

sergeyklay: I think this is Visual Studio specific stuff. For more see https://stackoverflow.com/questions/19024259/how-to-change-the-build-type-to-release-mode-in-cmake

skvadrik: Ah, ok. Good to know.

PR #317, opened 2020-10-09
GitHub actions improvements

sergeyklay:
Hello,

This PR amends the build matrix on GitHub Actions. The full used matrix is:

|  OS | Build Type  | Build Dir  | Common Flags | Skeleton |
| :---  |  ---  | --- | --- | --- |
| Linux | `Debug` | out-of-tree | `-O2 -fsanitize=undefined` | false |
| Linux | `Debug` | out-of-tree | `-O2 -fsanitize=address` | false |
| Linux | `Debug` | out-of-tree | `-O2` | false |
| Linux | `Debug` | in-tree | `-O2` | false |
| Linux | `Debug` | out-of-tree | `-O2` | **true** |
| Linux | `Release` | out-of-tree | — | false |
| macOs | `Debug` | out-of-tree | `-O2` | false  |
| macOs | `Debug` | out-of-tree | `-O2` | **true** |
| macOs | `Debug` |  in-tree | `-O2` | false  |
| macOs | `Release` | out-of-tree | — | false  |
| Windows | `Debug` | out-of-tree | — | false |
| Windows | `Debug` | in-tree | — | false  |
| Windows | `Release` | out-of-tree | — | false |

I removed `-g` for `Debug` configurations as it is used by default (see discussion in https://github.com/skvadrik/re2c/pull/316). There are still some things I need to port from Travis' configuration, but I'd like to do this in a separated PR. Let me know if I should amend this matrix.

Also I replaced deprecated `set-env` and `add-path` commands (see https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/ and [CVE-2020-15228](https://github.com/advisories/GHSA-mfwh-5m23-j46w)).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,15 +56,18 @@ jobs:
         run: |
           case ${{ runner.os }} in
             Linux*)
-              echo "::set-env name=CC::gcc"
-              echo "::set-env name=CXX::g++"
+              echo "CC=gcc" >> $GITHUB_ENV
+              echo "CXX=g++" >> $GITHUB_ENV
               ;;
             macOS*)
               sudo xcode-select -switch /Applications/Xcode.app
-              echo "::set-env name=CC::$(xcrun -f clang)"
-              echo "::set-env name=CXX::$(xcrun -f clang++)"
-              echo "::set-env name=SDKROOT::$(xcodebuild -version -sdk macosx Path)"
-              echo "::set-env name=PATH::$(dirname $(xcrun -f clang)):$PATH"
+
+              echo "CC=$(xcrun -f clang)" >> $GITHUB_ENV
+              echo "CXX=$(xcrun -f clang++)" >> $GITHUB_ENV
+              echo "SDKROOT=$(xcodebuild -version -sdk macosx Path)" >> $GITHUB_ENV
+
+              echo "$(xcodebuild -version -sdk macosx Path)" >> $GITHUB_PATH
+
               ;;
             Windows*)
               choco install --no-progress -y winflexbison3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,27 +16,101 @@ jobs:
 
     strategy:
       matrix:
-        toolchain:
-          - linux-gcc
-          - macos-clang
-          - windows-msvc
-
-        build-type:
-          - Debug
-          - Release
 
         include:
+          # Linux
           - toolchain: linux-gcc
             os: ubuntu-latest
-            compiler: gcc
+            build-type: Debug
+            build-dir: .build
+            common-flags: '-O2 -fsanitize=undefined'
+            skeleton: ${{ false }}
+
+          - toolchain: linux-gcc
+            os: ubuntu-latest
+            build-type: Debug
+            build-dir: .build
+            common-flags: '-O2 -fsanitize=address'
+            skeleton: ${{ false }}
+
+          - toolchain: linux-gcc
+            os: ubuntu-latest
+            build-type: Debug
+            build-dir: .build
+            common-flags: -O2
+            skeleton: ${{ false }}
+
+          - toolchain: linux-gcc
+            os: ubuntu-latest
+            build-type: Debug
+            build-dir: .
+            common-flags: -O2
+            skeleton: ${{ false }}
+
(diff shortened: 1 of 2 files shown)


skvadrik:
Awesome!

I have one question about `common-flags`: shouldn't linker use a different variable from compiler flags? Otherwise they end up with `-O2`, and it may get inconvenient later even if `-O2` is fine now. Linker flags should be empty for the most builds (except Asan and Ubsan).

Here is a wishlist, if you have time and energy for that:

- Build libre2c, re2go and documentation (CMake flags `-DRE2C_BUILD_RE2GO=on`, `-DRE2C_BUILD_LIBS=on` and `-DRE2C_REBUILD_DOCS=on`).

- Rebuild lexers: this requires a two-stage build. The first stage should build re2c in the minimal configuration, and the second stage should use the binary built at the first stage in `-DRE2C_FOR_BUILD` and set `-DRE2C_REBUILD_LEXERS=on`, and it should be a full build.

- Add a build with `-DCMAKE_CXX_FLAGS="-O2 -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC"`.

- Add Autotools build system.

- Why is it not possible to run tests on Windows? I think it should be fine in a Bash shell.

skvadrik:
(To clarify, I don't mean the wishlist for this PR.)

sergeyklay:
> I have one question about `common-flags`: shouldn't linker use a different variable from compiler flags?

Yes, this was a most strange stuff I'd like ask for. As I see in Travis' configuration the same flags are used for linker as well as compiler:
https://github.com/skvadrik/re2c/blob/8e56ce38af80f67421e45ab600601a135f8df0b1/.travis.yml#L29-L38
It looks like a mistake. So yes, only `CMAKE_CXX_FLAGS` is needed (and likely `CMAKE_C_FLAGS`).  Should I fix this in GitHub workflow?

About wishlist: yes, all the points from it are in my plans.

sergeyklay:
> Why is it not possible to run tests on Windows? I think it should be fine in a Bash shell.

I'd like to write a pure PowerShell wrapper for it a bit later and also add mingw support :)

sergeyklay:
By the way, today I thought about the fact that build using Ninja are usually faster. So, it would probably be nice to add this support too. What do you think about Ninja support? Should we just add yet another matrix item or just replace Make (used  by default by CMake)? We can use Ninja for any OS.

skvadrik:
> It looks like a mistake. So yes, only `CMAKE_CXX_FLAGS` is needed (and also perhaps `CMAKE_C_FLAGS`). Should I fix this in GitHub workflow?

Ah, it's actually my own mistake in travis.yml. Please correct it in the GitHub version. Note that you need to keep `-fsanitize=...` in `CMAKE_EXE_LINKER_FLAGS` and `CMAKE_SHARED_LINKER_FLAGS`.

> I'd like to write a pure PowerShell wrapper for it a bit later and also add mingw support :)

Nice! The run_test.sh.in script is quite horrible. It is too complex for bash and very inefficient. Most of the time is spend on sed and similar things. One day it will be rewritten in something like python.

> What do you think about Ninja support?

Can we enable if for some of the builds, and keep make for some others?

It might also be a good idea to build with bmake, to help portability on BSD. It's more restrictive than GNU make and sometimes the build breaks in unexpected ways.

sergeyklay:
Done. Take look at https://github.com/skvadrik/re2c/pull/317/commits/c90189f9e2d49de2840edfd769f9fefb2a02441d. I added some additional output to verify linker flags to be used. I hope they are OK now. We still don't say anything about which flags should be used for MSVC, but I'd like to do this work a bit later, in a separated PRs.

skvadrik:
Merged, thanks! I have one feature request: if possible, can you make the names of the builds (as they appear in left column in the GutHub UI) more descriptive? Right now one needs to click at each build to see the debug output (for example, there are many builds titled `Debug: linux-gcc`).

sergeyklay:
I want to make them a bit more clearer, but I can't think of a template yet. Do you have an idea?

skvadrik:
> Do you have an idea?

I'd probably just store a variable `name` in each build, and call them something like `linux-gcc-debug-ootree-skeleton`, `macos-clang-release-intree` and so on (manually hardcoding names in each build), if it's hard to come up with a good template.

PR #318, opened 2020-10-10
Added Asan and Ubsan CMake profiles

sergeyklay:
Hello,

This PR adds support of Asan and Ubsan build types for CMake. I prefer this option since it expresses the intent (run sanitizers) rather than modifies a number of flags. This also simplifies a bit build matrix on GitHub workflow.

Usage:
```bash
$ cmake \
    -DCMAKE_BUILD_TYPE=Asan \
    ...
    ...
```

Old school fans can still use the way with explicit passing of flags:
```bash
$ cmake \
    -DCMAKE_C_FLAGS="-g -fsanitize=address -fno-omit-frame-pointer" \
    -DCMAKE_CXX_FLAGS="-g -fsanitize=address -fno-omit-frame-pointer" \
    -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address" \
    -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address" \
    ...
    ...
```

Note, I added `-fno-omit-frame-pointer` to Asan build type and do not added `-O2` (it is added by workflow matrix). Let me know if you'd like to modify Asan/Ubsan build types by adding/removing flags.

### DIFF
--- a/cmake/Re2cBuildType.cmake
+++ b/cmake/Re2cBuildType.cmake
@@ -1,14 +1,98 @@
-set(_allowed_build_types Debug Release RelWithDebInfo MinSizeRel)
+# CMAKE_BUILD_TYPE is not used by multi-configuration generators.
+# Thus, we need to check this first.
+get_property(isMultiConfig GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
 
-# 1. The default value for CMAKE_BUILD_TYPE is an empty string.
-# 2. User can set CMAKE_BUILD_TYPE to any value at the cmake command line.
-#
-# Therefore, we check both cases, and make sure that we are dealing with
-# a known build type (if any).
-#
-# For more see discussion at https://github.com/skvadrik/re2c/pull/316
-if(CMAKE_BUILD_TYPE AND NOT CMAKE_BUILD_TYPE IN_LIST _allowed_build_types)
-    message(FATAL_ERROR "Invalid build type: ${CMAKE_BUILD_TYPE}")
+if(NOT isMultiConfig) # Makefiles, Ninja, ...
+    set(_allowed_build_types Asan Ubsan Debug Release RelWithDebInfo MinSizeRel)
+    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "${_allowed_build_types}")
+
+    # - The default value for CMAKE_BUILD_TYPE is an empty string
+    # - User can set CMAKE_BUILD_TYPE to any value at the cmake command line
+    #
+    # Therefore, we check both cases, and make sure that we are dealing with
+    # a known build type (if any).
+    #
+    # For more see discussion at https://github.com/skvadrik/re2c/pull/316
+    if(CMAKE_BUILD_TYPE AND NOT CMAKE_BUILD_TYPE IN_LIST _allowed_build_types)
+        message(FATAL_ERROR "Invalid build type: ${CMAKE_BUILD_TYPE}")
+    endif()
+
+    unset(_allowed_build_types)
+elseif(NOT MULTICONFIG_DONE) # Xcode, Visual Studio, ...
+    set(MULTICONFIG_DONE TRUE)
+    if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+        list(APPEND CMAKE_CONFIGURATION_TYPES Asan Ubsan)
+
+        # This is needed because user can set CMAKE_CONFIGURATION_TYPES
+        # at the cmake command line
+        list(REMOVE_DUPLICATES CMAKE_CONFIGURATION_TYPES)
+
+        set(CMAKE_CONFIGURATION_TYPES "${CMAKE_CONFIGURATION_TYPES}" CACHE STRING
+            "Semicolon separated list of supported configuration types." FORCE)
+    endif()
 endif()
 
-unset(_allowed_build_types)
+# Notes for Windows with Visual Studio:
+#
+# - Address Sanitizer (Asan) currently is under experimental stage
+# - Undefined Behavior Sanitizer (Ubsan) is not supported at this time
+#
+# Therefore, we have disabled support for these build types now, but
+# this support may be enabled in the future.
+#
+# Notes about '-fno-omit-frame-pointer' flag:
+#
+# Frame pointer omission does make debugging significantly harder. Local
+# vars are harder to locate and stack traces are much harder to reconstruct
+# without a frame pointer to help out. Also, accessing parameters can get more
+# expensive since they are far away from the top of the stack and may require
+# more expensive addressing modes.
+#
+# The '-fno-omit-frame-pointer' option direct the compiler to generate code
+# that maintains and uses stack frame pointer for all functions so that a
+# debugger can still produce a stack backtrace even with optimizations flags
+# (eg '-O1', '-O2', etc).
+if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+    # Setup Asan flags
+    set(CMAKE_C_FLAGS_ASAN
+        "${CMAKE_C_FLAGS_DEBUG} -fsanitize=address -fno-omit-frame-pointer" CACHE STRING
+        "Flags used by the C compiler for Asan build type or configuration." FORCE)
+
+    set(CMAKE_CXX_FLAGS_ASAN
+        "${CMAKE_CXX_FLAGS_DEBUG} -fsanitize=address -fno-omit-frame-pointer" CACHE STRING
+        "Flags used by the C++ compiler for Asan build type or configuration." FORCE)
+
+    set(CMAKE_EXE_LINKER_FLAGS_ASAN
+        "${CMAKE_EXE_LINKER_FLAGS_DEBUG} -fsanitize=address" CACHE STRING
+        "Linker flags to be used to create executables for Asan build type." FORCE)
+
+    set(CMAKE_SHARED_LINKER_FLAGS_ASAN
+        "${CMAKE_SHARED_LINKER_FLAGS_DEBUG} -fsanitize=address" CACHE STRING
+        "Linker flags to be used to create shared libraries for Asan build type." FORCE)
+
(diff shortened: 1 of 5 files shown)


cmake/Re2cBuildType.cmake
```diff
@@ -1,14 +1,76 @@
...
+
+if(isMultiConfig) # Xcode, Visual Studio, ...
+    if (NOT WIN32)
+        if(NOT "Asan" IN_LIST CMAKE_CONFIGURATION_TYPES)
+            list(APPEND CMAKE_CONFIGURATION_TYPES Asan)
+        endif()
+        if(NOT "Ubsan" IN_LIST CMAKE_CONFIGURATION_TYPES)
+            list(APPEND CMAKE_CONFIGURATION_TYPES Asan)
+        endif()
+    endif()
```
skvadrik, line 14: Can you change the structure of the code as follows, to simplify the handling of multi-config builds:
```cmake
if(NOT isMultiConfig) # Makefiles, Ninja, ...
    ...
else if(NOT MULTICONFIG_DONE) # Xcode, Visual Studio, ...
    set(MULTICONFIG_DONE TRUE)
    if (NOT WIN32)
        list(APPEND CMAKE_CONFIGURATION_TYPES Asan Ubsan)
    endif()
endif()
```
Also note that line 12 has `Asan` instead of `Ubsan`.

Is multi-configuration build automatically set depending on the platform, or is it something the user can choose as well?

cmake/Re2cBuildType.cmake
```diff
@@ -1,14 +1,76 @@
...
+    #
+    # For more see discussion at https://github.com/skvadrik/re2c/pull/316
+    if(CMAKE_BUILD_TYPE AND NOT CMAKE_BUILD_TYPE IN_LIST _allowed_build_types)
+        message(FATAL_ERROR "Invalid build type: ${CMAKE_BUILD_TYPE}")
+    endif()
+
+    unset(_allowed_build_types)
+endif()
+
+# AddressSanitizer (ASan) for Windows with MSVC is under experimental stage
```
skvadrik, line 39: Minor thing, but some comments spell "ASan" and some "Asan", can you unify that? Either way is fine by me. And similarly with Ubsan.

Should the comment mention Ubsan as well, since it's also disabled on Windows?

skvadrik (commented):
This is an elegant CMake way to deal with Asan/Ubsan builds. :)

Not to stop this PR, but a few thoughts I want to share.

A general recommendation that I have is to move all complexity out of the build system and to various build scripts, CI configs and so on. In the future re2c may move to a different hosting, or to a different CI, but the build system lives as long as the project lives, and the fewer gory details it has, the better.

One of the things about CMake is that it allows a lot of flexibility, because it has conditional statements and loops in the configuration part of the build system. It is convenient, but it also means that it is easy to write a lot of code, which is always a double-sided sword.

Also I think it's important to keep Autotools on par with CMake when possible.

cmake/Re2cBuildType.cmake
```diff
@@ -1,14 +1,76 @@
...
+        if(NOT "Ubsan" IN_LIST CMAKE_CONFIGURATION_TYPES)
+            list(APPEND CMAKE_CONFIGURATION_TYPES Asan)
+        endif()
+    endif()
+else() # Makefiles, Ninja, ...
+    if (NOT WIN32)
+        set(_allowed_build_types Asan Ubsan Debug Release RelWithDebInfo MinSizeRel)
+    else()
+        set(_allowed_build_types Debug Release RelWithDebInfo MinSizeRel)
+    endif()
```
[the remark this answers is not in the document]
sergeyklay: Did you mean:

```cmake
list(APPEND _allowed_build_types Asan Ubsan)
```

There is no `append` CMake command.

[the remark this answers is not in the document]
skvadrik: Yes, that's right.

skvadrik:
Can you squash the commits for this PR into one before merging?

sergeyklay:
> A general recommendation that I have is to move all complexity out of the build system and to various build scripts, CI configs and so on. In the future re2c may move to a different hosting, or to a different CI, but the build system lives as long as the project lives, and the fewer gory details it has, the better.

Does this mean that you'd like to get building tools dispersed across various build scripts as well as CI configuration instead of a centralized and specialized tool? IMO, having a vendor lock on the CI configuration isn't a forward-looking strategy. On the other hand, the project build logic scattered across various scripts doesn't help to get rid of build errors and overall project complexity. And this design makes project depend on the particular OS/shell. I'm sorry, however I didn't fully understand your point here. Could you please elaborate on that a bit more?

> One of the things about CMake is that it allows a lot of flexibility, because it has conditional statements and loops in the configuration part of the build system. It is convenient, but it also means that it is easy to write a lot of code, which is always a double-sided sword.

CMake looks like more cross-platform than, for example Autotools or Bash. This means that you are likely will get fewer issues on various systems, including BSD, Windows, macOS and so on. So I'd like to know more about the double-sided sword :)

sergeyklay:
> Also I think it's important to keep Autotools on par with CMake when possible.

I'll try add Asan and Ubsan for Autotools a bit later.

skvadrik:
> Does this mean that you'd like to get building tools dispersed across various build scripts as well as CI configuration instead of a centralized and specialized tool?

I think that CI, maintainers/developers and users have slightly different needs. Users need a very simple and standard way to build the project, preferably documented in README (`configure && make && make install` for Autotools, or `cmake . -B $BUILDDIR && make` for CMake). Maintainers/developers need flexibility, such as the ability to enable/disable parts of the project, do bootstrap builds or enable certain build flags. And CI just needs to be able to build the project in different configurations in some automated way.

I think that build system should not be cluttered with details related to VCS, hosting, CI and other infrastructural details, in the sense that should be possible to migrate any of these details without rewriting the build system. Build scripts are for developers' convenience, so they don't have to be portable and can afford some lousiness (they can be easily written or removed). CI configs are for automated testing, they can be relatively complex or cluttered with details.

Documentation and READMEs should reference only the build system, not build scrips or CI.
 
> IMO, having a vendor lock on the CI configuration isn't a forward-looking strategy.

I very much agree with this, therefore I don't think that GitHub Actions can be a canonical reference of how to build re2c. On one hand, it's an infrastructural detail and can go away in the future (I hope not, but there are sad examples). On the other hand, it is too complex and cluttered with details to serve as a good reference of how to build the project (and it requires understanding of the special syntax).

> On the other hand, the project build logic scattered across various scripts doesn't help to get rid of build errors and overall project complexity. And this design makes project depend on the particular OS/shell.

I never thought of the build scripts as anything but a convenience for developers. Sure, they are not very clean or portable, but they can be added or thrown away at any time without much consideration.

> CMake looks like more cross-platform than, for example Autotools or Bash. This means that you are likely will get fewer issues on various systems, including BSD, Windows, macOS and so on. So I'd like to know more about the double-sided sword :)

True, CMake has this great advantage. But as a build system, it lacks clear separation between the configuration part and the scripting part of the build system. It allows to write a lot of code easily, with many conditionals and complex logic, and that should be used with caution.

But this is only my vision, I may be wrong and I don't want to enforce it on the project. Since you contribute so much, I can just give you commit access. Then you'll be able to implement your ideas more freely and just ask me for advice if necessary. What do you think?

sergeyklay:
Well, as you mentioned there are three "consumers" groups:

**A.** Regular users
**B.** Maintainers / developers
**C.** CI/CD systems

I think my efforts at this time are aimed for **C** and partially for **B**. I believe that proper CMake design will not interfere for **A** and will help for **B** and **C**. I'm trying to make interop with CI a little simpler and clearer, this is why some CMake changes are needed. In the end I would like to get a portable build system without dependencies on CI platform. Let's look at what I'm doing from a different side. The fact that we are getting more and more powerful build tools does not interfere (I hope it helps) any of the consumers groups listed.

I am impressed by the idea of ​​sending the user to read the CI configuration, especially when the documentation isn't comprehensive. May be because this is an easiest and short way. But I absolutely agree with you that documentation should reference only the build system and this is what I planed to ask you in the future -- create a complete and comprehensive documentation on how to build project using CMake, various OSes and supported build options. And I'll be happy to help with this. We should probably do the same for Autotools as long as it planned for long term support. To summarize -- lack of clear and comprehensive documentation should not be a reason for rejecting a build  tool or good practices.

I’d be happy to have commit access, but don’t think you’ll get rid of me so easily haha. I will still need a review for my PRs and your opinion on that.

cmake/Re2cBuildType.cmake
```diff
@@ -1,14 +1,85 @@
...
+        "${CMAKE_EXE_LINKER_FLAGS_DEBUG} -fsanitize=address -fno-omit-frame-pointer" CACHE STRING
+        "Linker flags to be used to create executables for Asan build type." FORCE)
+
+    set(CMAKE_SHARED_LINKER_FLAGS_ASAN
+        "${CMAKE_SHARED_LINKER_FLAGS_DEBUG} -fsanitize=address -fno-omit-frame-pointer" CACHE STRING
+        "Linker flags to be used to create shared libraries for Asan build type." FORCE)
+
+    set(CMAKE_STATIC_LINKER_FLAGS_ASAN
+        "${CMAKE_STATIC_LINKER_FLAGS_DEBUG} -fsanitize=address -fno-omit-frame-pointer" CACHE STRING
+        "Linker flags to be used to create static libraries for Asan build type." FORCE)
```
skvadrik, line 62: Is `-fno-omit-frame-pointer` needed for linker?

skvadrik (commented):
> I’d be happy to have commit access

Granted. Welcome to the club! \o/

> but don’t think you’ll get rid of me so easily haha. I will still need a review for my PRs and your opinion on that

Sure, just add me to review, I'm always happy to discuss.

cmake/Re2cBuildType.cmake
sergeyklay: I've updated this part - added notes and removed linker flag. Please take a look.

skvadrik (approved):

skvadrik:
Approved, I will leave it for you to click the merge button. :)

sergeyklay:
Thank you for sharing your opinion, help with the review, and for helping me making this feature better.

Issue #302, opened 2020-08-19
Using CMake add_subdirectory fails


hanak:
Before I get to the description, I appreciate the fact that you have added support for CMake. It helped me with the multi-platform environment a lot. Thank you for doing so. Yet, let us get back to the behaviour I observed.

When I try to use `add_subdirectory("<path to re2c source code outside source tree>" "<path to binary dir>")`, CMake 3.13.x (the version I have to use) fails with:
```
CMake Error at external/re2c-2.0.2/CMakeLists.txt:127 (add_library):
  Cannot find source file:

    <binary dir on my machine>/src/msg/help.cc

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx


CMake Error at external/re2c-2.0.2/CMakeLists.txt:127 (add_library):
  No SOURCES given to target: re2c_objects_autogen
```

I noticed that re2c/CMakeLists.txt contains at the line 234 following:
`if (CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)`

When I disable this condition (e.g. `if (CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)`), build works. Why is the condition there?

Also, I have noticed that the line 311 excludes non-UNIX machines from installation through `if(UNIX)`. This means, if I want to deploy the build output somewhere using Windows machine I have to do it on my own while on Linux (and cross-compiling for Windows on Linux), re2c's CMakeLists.txt will install it. It this intentional? I am sorry for adding another issue into this one. I can separate it if there is no straight answer.

Context:
I am trying to build re2c as a prerequisite of another project. The build has to run automatically on a build machine in order to allow CI/CD pipeline and it should run on a developer's machine, too. Developer's machine might be either Linux or Windows.

skvadrik:
Hi @hanak ! think both `if (CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)` and `if(UNIX)` conditions were added by @ligfx in the [initial pull request](https://github.com/skvadrik/re2c/pull/275/commits/a530570e5164ae302f89b79f9e6b04032c697245). Perhaps @ligfx can explain why they are needed (if they are indeed needed). I think @ligfx intended to make it possible to build re2c as a subproject.

Meanwhile, can you provide a minimal example of your top-level CMakeLists.txt in which you try to add re2c as a subproject?

[...]

PR #319, opened 2020-10-13
Add minimal install test on GitHub Actions

sergeyklay:
This PR adds ability to install re2c as well as re2go on GitHub Actions.

Another notable changes are:

#### Install on Windows

I've started work to enable `install` targets for Windows. There is no ability to install libs yet, but the main binaries can be already installed.

#### Minimal Install Test

Renamed `Minimal Load Test (Windows)` job to more generic  `Minimal Install Test`. We'll amend this job in the future. But right now it tests that after installation the necessary binaries are in the expected place and you can run them (on Windows too).

#### Jobs Naming

Make the names of the builds more descriptive as discussed here:  https://github.com/skvadrik/re2c/pull/317#issuecomment-706685657

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,7 @@ defaults:
 
 jobs:
   build-and-test:
-
-    name: "${{ matrix.build-type }}: ${{ matrix.toolchain }}"
+    name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
 
     strategy:
@@ -19,7 +18,7 @@ jobs:
 
         include:
           # Linux
-          - toolchain: linux-gcc
+          - name: linux-gcc-ubsan-ootree
             os: ubuntu-latest
             # Undefined Behavior Sanitizer (cmake -DCMAKE_BUILD_TYPE=Ubsan)
             build-type: Ubsan
@@ -27,7 +26,7 @@ jobs:
             cxx-flags: '-O2'
             skeleton: ${{ false }}
 
-          - toolchain: linux-gcc
+          - name: linux-gcc-asan-ootree
             os: ubuntu-latest
             # Address Sanitizer (cmake -DCMAKE_BUILD_TYPE=Asan)
             build-type: Asan
@@ -35,28 +34,28 @@ jobs:
             cxx-flags: '-O2'
             skeleton: ${{ false }}
 
-          - toolchain: linux-gcc
+          - name: linux-gcc-debug-ootree
             os: ubuntu-latest
             build-type: Debug
             build-dir: .build
             cxx-flags: -O2
             skeleton: ${{ false }}
 
-          - toolchain: linux-gcc
+          - name: linux-gcc-debug-intree
             os: ubuntu-latest
             build-type: Debug
             build-dir: .
             cxx-flags: -O2
             skeleton: ${{ false }}
 
-          - toolchain: linux-gcc
+          - name: linux-gcc-debug-ootree-skeleton
             os: ubuntu-latest
             build-type: Debug
             build-dir: .build
             cxx-flags: -O2
             skeleton: ${{ true }}
 
-          - toolchain: linux-gcc
+          - name: linux-gcc-release-ootree
             os: ubuntu-latest
             build-type: Release
             build-dir: .build
@@ -64,28 +63,28 @@ jobs:
             skeleton: ${{ false }}
 
           # macOS
-          - toolchain: macos-clang
+          - name: macos-clang-debug-ootree
             os: macos-latest
             build-type: Debug
             build-dir: .build
             cxx-flags: -O2
             skeleton: ${{ false }}
 
-          - toolchain: macos-clang
+          - name: macos-clang-debug-ootree-skeleton
             os: macos-latest
             build-type: Debug
             build-dir: .build
             cxx-flags: -O2
             skeleton: ${{ true }}
 
-          - toolchain: macos-clang
+          - name: macos-clang-debug-intree
             os: macos-latest
             build-type: Debug
             build-dir: .
             cxx-flags: -O2
             skeleton: ${{ false }}
 
-          - toolchain: macos-clang
+          - name: macos-clang-release-ootree
             os: macos-latest
             build-type: Release
             build-dir: .build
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -155,13 +158,15 @@ jobs:
             Windows*)
               cmake -S . \
                 -B ${{ matrix.build-dir }} \
+                -DCMAKE_INSTALL_PREFIX="$(pwd)/install" \
                 -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} \
```
skvadrik, line 162: This can be factored out of the platform switch.

.github/workflows/ci.yml
```diff
@@ -191,3 +191,14 @@ jobs:
         if: startsWith(runner.os, 'Windows') == false &&  matrix.skeleton == true
         working-directory: ${{ matrix.build-dir }}
         run: ./run_tests.sh --skeleton
+
+      - name: Install
+        run: |
+          cmake --build ${{ matrix.build-dir }} --target install
+          find install
```
skvadrik: What does this `find install` do? Can't find it in GitHub Actions docs. :)

CMakeLists.txt
```diff
@@ -287,15 +287,13 @@ add_custom_target(docs DEPENDS "${re2c_docs}")
...
-            install(FILES "${re2c_manpage_go}" DESTINATION "share/man/man1")
-        endif()
-        install(FILES include/unicode_categories.re DESTINATION "${RE2C_STDLIB_DIR}")
+    install(TARGETS re2c RUNTIME DESTINATION bin)
+    install(FILES "${re2c_manpage_c}" DESTINATION "share/man/man1")
+    if(RE2C_BUILD_RE2GO)
+        install(TARGETS re2go RUNTIME DESTINATION bin)
+        install(FILES "${re2c_manpage_go}" DESTINATION "share/man/man1")
     endif()
+    install(FILES include/unicode_categories.re DESTINATION "${RE2C_STDLIB_DIR}")
```
skvadrik, line 296: To me this change seems correct, but when the question of Windows install rules came up in a discussion some time ago (https://github.com/skvadrik/re2c/issues/302#issuecomment-676672529), there were concerns about default install paths (which I don't understand, but I don't have Windows).

It works in the CI environment ([here](https://github.com/skvadrik/re2c/pull/319/checks?check_run_id=1247973854)), but uses the POSIX file hierarchy with forward slashes:
```
  -- Install configuration: "Debug"
  -- Installing: D:/a/re2c/re2c/install/bin/re2c.exe
  -- Installing: D:/a/re2c/re2c/install/share/man/man1/re2c.1
  -- Installing: D:/a/re2c/re2c/install/bin/re2go.exe
  -- Installing: D:/a/re2c/re2c/install/share/man/man1/re2go.1
  -- Installing: D:/a/re2c/re2c/install/share/re2c/stdlib/unicode_categories.re
```
And the choice of `D:/a/re2c` prefix is unclear to me: where does it come from?

On the whole, I'd rather enable install rules on Windows and wait until someone complains, but I'm far from understanding how the Windows environment works.

.github/workflows/ci.yml
```diff
@@ -19,95 +22,95 @@ jobs:
 
         include:
           # Linux
-          - toolchain: linux-gcc
+          - toolchain: linux-gcc-out-of-tree
```
skvadrik: This and all similar changes seem unnecessary in this commit (especially since they are removed by the following https://github.com/skvadrik/re2c/pull/319/commits/9dbcee51b5f7ec49effd0c3a0fc0aca92a9e0704), can you remove them?

.github/workflows/ci.yml
```diff
@@ -188,14 +188,14 @@ jobs:
...
+        if: startsWith(runner.os, 'Windows') == false && matrix.skeleton == true
         working-directory: ${{ matrix.build-dir }}
         run: ./run_tests.sh --skeleton
 
       - name: Install
         run: |
-          cmake --build ${{ matrix.build-dir }} --target install
-          find install
+          cmake --build ${{ matrix.build-dir }} --config ${{ matrix.build-type }} --target install
+          echo -e "Intall tree:\n$(find install | sed 's/^/-- /')"
```
skvadrik, line 198: This is part of the previous commit https://github.com/skvadrik/re2c/pull/319/commits/986af5bebb0552d5a39f825385eb73349ab977bf, right? Can you move it there?

Also, there is a small typo: `Intall` (missing "s").

skvadrik (changes requested):
Nice! Now it's easy to see which build is where.

skvadrik (changes requested):
Install test is very useful!

.github/workflows/ci.yml
sergeyklay: There is no `tree`, so `find install` just prints `install` directory recursively

CMakeLists.txt
sergeyklay: `D:/a/re2c/re2c` is a project directory inside CI environment . I have Windows and I see the same result / output

.github/workflows/ci.yml
sergeyklay: I can't deal with ` -A x64` right now. Only this (which is needed only for Visual Studio) stops me from having one common 'Configure' for all generators. Adding `-A x64` for all generators leads to CMake error. I'll try to sort out with 'Configure' a bit later

.github/workflows/ci.yml
sergeyklay: I'll squash commits at the end. Is this fits?

.github/workflows/ci.yml
sergeyklay: I'll squash commits at the end.  Thank you for the missing 's'!

.github/workflows/ci.yml
skvadrik: I didn't mean a common configure step, only to have a shell variable before the `switch` that stores common cmake options (since you already have an array `args` for non-Windows case, it can be initialized before `switch`).

.github/workflows/ci.yml
skvadrik: Ah! I forgot this line is part of  shell script, and assumed it's some special GitHub Actions thing. My bad.

.github/workflows/ci.yml
skvadrik: Sure, although I think it's better to keep unrelated changes as separate commits. But squashing is also fine, as long as the commit message lists the individual changes.

CMakeLists.txt
skvadrik: I think the default prefix on Windows is `c:/Program Files/${PROJECT_NAME}`, at least that's what [CMake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html) says. I'm not suggesting to override it.

skvadrik (approved):

PR #320, opened 2020-10-15
Fixed '-Wunreachable-code' warning

sergeyklay:
Hello,

This code will never be executed (was found by Clang):

```text
../src/parse/lex.re:428:9: warning: code will never be executed
      [-Wunreachable-code]
        goto scan;
        ^~~~~~~~~
```

### DIFF
--- a/src/parse/lex.re
+++ b/src/parse/lex.re
@@ -425,7 +425,6 @@ scan:
 
     * {
         msg.fatal(tok_loc(), "unexpected character: '%c'", *tok);
-        goto scan;
     }
 */
 }


skvadrik:
When changing re2c source code written in re2c (generally any source files that have .re extension), it is necessary to rebuild the precompiled files in the bootstrap directory. These files are used to build re2c without an already existing re2c executable.
```
$ mkdir stage1 && cd stage1
$ cmake .. && make
$ cd .. && mkdir stage2 && cd stage2
$ cmake .. -DRE2C_REBUILD_LEXERS=yes -DRE2C_FOR_BUILD=$(pwd)/../stage1/re2c && make
```
This will update all bootstrap lexers but for the most part the only change will be an updated generation date (this should be reverted). There will be one truly updated bootstrap file, bootstrap/src/parse/lex.cc, that should be committed together with src/parse/lex.re. It is not necessary to build stage1 if you have a fresh enough (and stable enough) re2c elsewhere on your system.

Regarding the change itself, I'm a bit concerned that not all compilers will see that `fatal` is `noreturn`. It is specified as an attribute on the function declaration in a header file, but `RE2C_ATTR` is conditionally defined under `__GNUC__`. So I'd expect that Visual Studio wouldn't be able to see that `fatal` does not return. And the control flow after `fatal` depends on where re2c puts this semantic action. Also, it is really easy to change `fatal` to, say, `error` (which is like `fatal` but returns) and forget to add `goto` or `return`. 

The same problem is raised from time to time by re2c users: it would be good to have a dead `goto` or `assert(false)` or some similar statement after semantic actions that would serve as a precaution, but not trigger `-Wunreachable-code` warnings.

sergeyklay:
> Regarding the change itself, I'm a bit concerned that not all compilers will see that `fatal` is `noreturn`. It is specified as an attribute on the function declaration in a header file, but `RE2C_ATTR` is conditionally defined under `__GNUC__`.

In fact, I am impressed by the idea of making this macro a bit more portable. For example Clang supports
`__attribute__((noreturn))` and Visual Studio supports `__declspec(noreturn)`. I have never applied this widely. However, I am curious if a more generic implementation is possible.

> it is really easy to change `fatal` to, say, `error` (which is like `fatal` but returns) and forget to add `goto` or `return`.

Well, only this file has 22 places with `fatal`  without any dead code after. For example:

https://github.com/skvadrik/re2c/blob/a904ff8855e0d717625f0fe3740387a914adaf4b/src/parse/lex.re#L255-L257

https://github.com/skvadrik/re2c/blob/a904ff8855e0d717625f0fe3740387a914adaf4b/src/parse/lex.re#L278-L280

https://github.com/skvadrik/re2c/blob/a904ff8855e0d717625f0fe3740387a914adaf4b/src/parse/lex.re#L340-L342

So, to be more precise, my change affected the only usage of `fatal` where dead code is used. All other occurrences don't use such approach. Should we take a similar approach everywhere and add dead code since it is part of the design?

sergeyklay:
From conceptual point of view I'd like propose to drop this function at all and use `error` everuvery with an explicit `exit(1)`. It seems to me doubtful to introduce into the design such an obscure thing that can mislead not only an user, but also the compiler, just to save 1 line of code. Explicit is better than implicit.

skvadrik:
> I am impressed by the idea of making this macro a bit more portable.

Yes, that would be good. Initially I wanted some quick and dirty hack to have some validation with GCC, but there is no reason not to make it more portable.

> Well, only this file has 22 places with fatal without any dead code after.

Good point, I don't like all 22. :D 

> In fact, it looks like fatal should never have interrupted the main thread and `exit(1)` was added by mistake because a quick inspection showed about 60 use cases across src/ w/o dead code :)

No, I don't think it was added accidentally, previously there was only `fatal` that exits, but no `error`.

> From conceptual point of view I'd like propose to drop this function at all and use `error` everywhere with an explicit `exit(1)`.

I think this is the best way. A bit more work, a bit more code, but no assumptions and workarounds.

In the future I'd like to have better error handling without `exit(1)` in the middle of a program, as this doesn't clean up resources and makes Valgrind tests fail.

sergeyklay:
> I think this is the best way. A bit more work, a bit more code, but no assumptions and workarounds.

So, should I continue the work was started in this PR? Or will be better close this and create a new one for replacing `fatal` for `error`?

sergeyklay:
> No, I don't think it was added accidentally, previously there was only `fatal` that exits, but no `error`.

BTW, from my point of view its more about C-style, not C++ one. I mean it is is a common (albeit perverse) approach of doing things in the C world :)

skvadrik:
> So, should I continue the work was started in this PR? Or will be better close this and create a new one for removing fatal in favor of error & exit(1)?

I think the second option is better (removing `fatal`).

> BTW, from my point of view its more about C-style, not C++ one. I mean it is a common (albeit perverse) approach of doing things in the C world :)

Do you mean that explicit `exit(1)` without freeing resources is C style?

sergeyklay:
> When changing re2c source code written in re2c (generally any source files that have .re extension), it is necessary to rebuild the precompiled files in the bootstrap directory. These files are used to build re2c without an already existing re2c executable.
> 
> ```
> $ mkdir stage1 && cd stage1
> $ cmake .. && make
> $ cd .. && mkdir stage2 && cd stage2
> $ cmake .. -DRE2C_REBUILD_LEXERS=yes -DRE2C_FOR_BUILD=$(pwd)/../stage1/re2c && make
> ```

It is interesting that this does not work at all. Any combination of cmake command at stage2 leads to an error:
```
$ make
[  1%] Generating src/msg/ver_to_vernum.cc
../src/msg/ver_to_vernum.re:33:19: warning: tag 'p2' has 2nd degree of nondeterminism [-Wnondeterministic-tags]
../src/msg/ver_to_vernum.re:33:19: warning: tag 'e2' has 2nd degree of nondeterminism [-Wnondeterministic-tags]
../src/msg/ver_to_vernum.re:33:19: warning: tag 'p3' has 2nd degree of nondeterminism [-Wnondeterministic-tags]
../src/msg/ver_to_vernum.re:33:19: warning: tag 'e3' has 2nd degree of nondeterminism [-Wnondeterministic-tags]
re2c: error: cannot open output file src/msg/ver_to_vernum.cc
make[2]: *** [CMakeFiles/re2c_objects_autogen_ver_to_vernum.dir/build.make:62: src/msg/ver_to_vernum.cc] Error 1
make[1]: *** [CMakeFiles/Makefile2:155: CMakeFiles/re2c_objects_autogen_ver_to_vernum.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

just because `cmake .. -DRE2C_REBUILD_LEXERS=yes -DRE2C_FOR_BUILD=$(pwd)/../stage1/re2c && make` does not create `src/` directory inside `stage2/` (but `cmake ..` at stage1 does). The same behavior with:

- ` cmake .. -DRE2C_REBUILD_LEXERS=yes -DRE2C_FOR_BUILD=/usr/bin/re2c`
- ` cmake .. -DRE2C_REBUILD_LEXERS=yes -DRE2C_FOR_BUILD=$(pwd)/../stage1/re2c`
- ` cmake -S .. -B . -DRE2C_REBUILD_LEXERS=yes -DRE2C_FOR_BUILD=$(pwd)/../stage1/re2c`
- ` cmake -S .. -B . -DRE2C_REBUILD_LEXERS=yes -DRE2C_FOR_BUILD=/usr/bin/re2c`

generating a `stage2/` directory contents using any of these methods results in the `stage2/src/` directory not being created and we get `re2c: error: cannot open output file src/msg/ver_to_vernum.cc`.

skvadrik:
Hmm, I see, there is a problem with doing the `cmake .. -DRE2C_REBUILD_LEXERS=yes -DRE2C_FOR_BUILD=$(pwd)/../stage1/re2c && make` for the first time. I normally don't see it because I reconfigure in an already existing directory. You can use a workaround and do the first stage2 build without `RE2C_REBUILD_LEXERS`, and the second one with it:
```
$ mkdir stage1 && cd stage1
$ cmake .. && make
$ cd .. && mkdir stage2 && cd stage2
$ cmake .. && make
$ #... make your changes
$ cmake .. -DRE2C_REBUILD_LEXERS=yes -DRE2C_FOR_BUILD=$(pwd)/../stage1/re2c && make
```
The problem with first build needs to be fixed, but it's a separate problem.

skvadrik:
I have a patch to fix the issue with the first boot: https://github.com/skvadrik/re2c/pull/322.

sergeyklay:
After re-generating the bootstrap folder (using my own workaround) and force-pushed to my repo I have 1 failed test: https://github.com/sergeyklay/re2c/runs/1268151861

Do you have ant idea? The diff is https://github.com/skvadrik/re2c/compare/master...sergeyklay:drop-message-fatal?expand=1

skvadrik:
I don't see any obvious reason for that test to fail. Did you run the tests locally after applying your patch? If any tests fail, there will be a temporary test directory with the failed ones, where you can see all outputs and compare the expected result with the actual one. Can you convert your patch into a PR, so that I can test it locally?

Also, it is important to make sure that your `RE2C_FOR_BUILD` was build from a clean source tree that passes all the tests. With self-hosting programs it is easy to get into a vicious feedback loop when the error propagates from one stage to another.

sergeyklay:
OK, lets merge #322 and I'll try again

skvadrik:
Ah, I see it. You forgot an `exit(1)` at line 448 of src/regexp/ast_to_re.cc. I found it by looking at the error message text in the failed test, and searching for it in the diff.

sergeyklay:
> Ah, I see it. You forgot an `exit(1)` at line 448 of src/regexp/ast_to_re.cc. I found it by looking at the error message text in the failed test, and searching for it in the diff.

Yay! Thank you very much!

skvadrik:
One thing I just noticed is that GitHub "Merge pull request" button now has variants in the drop-down menu. I have a suggestion: let's always use "Rebase and merge" option, ok? I want the history to be cluttered with merge commits.

sergeyklay:
> One thing I just noticed is that GitHub "Merge pull request" button now has variants in the drop-down menu.

Oh! Just FYI, this feature is from 2016 https://github.blog/2016-09-26-rebase-and-merge-pull-requests/ :) 

> I have a suggestion: let's always use "Rebase and merge" option, ok? I want the history to be cluttered with merge commits.

I completely agree with you! Lets use "Rebase and Merge". Btw, How long have you been looking [at the settings](https://github.com/skvadrik/re2c/settings)?

image

skvadrik:
> Just FYI, this feature is from 2016

Eh... and I always bothered with command-line instructions. :D 

> Btw, How long have you been looking at the settings?

A while ago. :) Thank for reminding, I checked box "Allow rebase merging" and "Automatically delete head branches". Good stuff.

PR #321, opened 2020-10-16
Fixed linker flags for Asan and Ubsan profiles.

sergeyklay:
This is a fix for linker flags that were mistakenly added after discussion at https://github.com/skvadrik/re2c/pull/318.

Here I fell into the trap of terminology, because although STATIC_LINKER is technically a program for "getting one library from a bundle of something", it works *completely* differently than EXE_LINKER and SHARED_LINKER.

Right now I'm preparing another (unrelated) PR and this issue was discovered during linking static libraries.


skvadrik (commented):
The fix is good. I have a few suggestions regarding commit message (they apply to all commits in general):

- Can you use present tense (like `Fix` instead of `Fixed`)?
- Can you avoid uncertain terms like "fix" and instead try to tell what the fix is about? (For example, `Don't add CMAKE_STATIC_LINKER_FLAGS flags to Asan/Ubsan builds.`, with a bit of text further down explaining why you decided to drop them).

sergeyklay:
> I have a few suggestions regarding commit message (they apply to all commits in general):

Oh, no problem! I'm sorry, it wasn't obvious to me :-/ And of course your comments are clear to me, although I must admit I have often met the opposite point of view about commit patterns. Nevertheless, Is there any other rules? How about adding a few lines to CONTRIBUTTING.md? It definitely will help contributors in the future. What about PR subject? Is there any rule?

skvadrik (commented):
Some more bikeshedding from me: please don't use `Here I ...`, as it looks a bit out of the context. Just limit the sentence to `Although STATIC_LINKER is technically a program for "getting one library from a bundle of something", it works *completely* differently than EXE_LINKER and SHARED_LINKER.`.

skvadrik:
No worries, the rules are not set in stone, it's just something I think makes sense. But it's a good idea to have guidelines for contributors. Hopefully we can save the effort and borrow them from some other project, but here are a few guidelines for a start:

- Commit headline should use present tense, start with capital letter, end with period.
- Commit headline should be specific (avoid generic uncertain terms like "fix", "refactoring", "code cleanup", etc. when possible).
- Long commit message bodies with explanations, examples, test instructions are welcome.
- Unrelated changes should be split into multiple commits.

... feel free o suggest other good practices.

sergeyklay:
> * ommit headline should be specific (avoid generic uncertain terms like "fix", "refactoring", "code cleanup", etc. when possible).

There is an example http://whatthecommit.com/ (refresh page) :D

sergeyklay:
> ... feel free o suggest other good practices.

There is a good reference: https://www.conventionalcommits.org/en/v1.0.0/

skvadrik (approved):

PR #323, opened 2020-10-17
Drop 'fatal' and 'Msg::fatal' in favor of 'error' and 'Msg::error'

sergeyklay:
This is aimed to reduce assumptions, workarounds and potential misunderstandings due to not a completely obvious behavior of
'fatal' function. I hope that overall this will make the code a little more understandable (albeit a little more verbose).

Closes https://github.com/skvadrik/re2c/pull/320


.github/workflows/ci.yml
```diff
@@ -182,7 +182,7 @@ jobs:
 
       - name: Standard Tests
         if: startsWith(runner.os, 'Windows') == false &&  matrix.skeleton == false
-        run: cmake --build ${{ matrix.build-dir }} --target check
+        run: cmake --build ${{ matrix.build-dir }} --target check --verbose
```
skvadrik, line 185: This is an unrelated change. Not necessary for this PR, but it would be better to do it in a separate commit.

src/parse/validate.cc
```diff
@@ -41,28 +44,33 @@ void validate_ast(const specs_t &specs, const opt_t *opts, Msg &msg)
...
                 "code to default rule %sis already defined at line %u",
                 incond(i->name).c_str(), i->defs[0]->loc.line);
+            exit(1);
         }
         if (!i->eofs.empty() && opts->eof == NOEOF) {
-            msg.fatal(i->eofs[0]->loc,
+            msg.error(i->eofs[0]->loc,
                 "%sEOF rule found, but 're2c:eof' configuration is not set",
                 incond(i->name).c_str());
+          exit(1);
```
skvadrik, line 56: Is it misaligned?

src/regexp/ast_to_re.cc
```diff
@@ -251,7 +253,7 @@ RE *ast_to_re(RESpec &spec, const AST *ast, size_t &ncap, int32_t height)
 
         case AST::REF:
             if (misuse_of_named_def(spec, ast)) {
-                return NULL;
+                exit(1);
```
skvadrik, line 256: Don't change this, some day `misuse_of_named_def` will return a proper error, and the caller will need to check it and return up the stack. If you want to clean it up for now, replace `misuse_of_named_def` with a void function `check_misuse_of_named_def` and remove the `if`.

src/regexp/ast_to_re.cc
```diff
@@ -362,7 +367,7 @@ Range *ast_to_range(RESpec &spec, const AST *ast)
             if (spec.opts->posix_syntax) break;
             return ast_to_range(spec, ast->cap);
         case AST::REF:
-            if (misuse_of_named_def(spec, ast)) return NULL;
+            if (misuse_of_named_def(spec, ast)) exit(1);
```
skvadrik, line 370: Same as lines 254-256.

skvadrik (changes requested):
I don't see bootstrap parser being regenerated (it should, because parser.ypp has changed). Do you have bison? If not, can you get it?

src/regexp/ast_to_re.cc
sergeyklay: Well, this is how re2c works right now. Do you remember, `fatal` calls `exit(1)`?

.github/workflows/ci.yml
sergeyklay: Good catch!

src/regexp/ast_to_re.cc
sergeyklay: Added here and below

```c
                /* unreachable until 'misuse_of_named_def' uses 'exit(1)' */
                return NULL;
```

src/regexp/ast_to_re.cc
sergeyklay: Added here and above

```c
                /* unreachable until 'misuse_of_named_def' uses 'exit(1)' */
                return NULL;
```

sergeyklay:
> I don't see bootstrap parser being regenerated (it should, because parser.ypp has changed). Do you have bison? If not, can you get it?

Well, I reverted these changes on my local machine and did not commit them :( This was done on purpose because you didn't mention them on https://github.com/skvadrik/re2c/pull/320#issuecomment-709617201

I have a more recent bison version (3.7.1) that creates a huge diff. Is this acceptable?

src/regexp/ast_to_re.cc
skvadrik: I think it's better to change it like this (pseudo-diff):
```diff
- if (misuse_of_named_def(spec, ast)) return NULL;
+ check_misuse_of_named_def(spec, ast);
```
And the function definition:
```diff
- bool misuse_of_named_def(RESpec &spec, const AST *ast)
+ void misuse_of_named_def(RESpec &spec, const AST *ast)
```
Having a comment is way too much hassle for such a minor thing, the code is already complex by itself.

skvadrik:
> Well, I reverted these changes on my local machine and did not commit them :( This was done on purpose because you didn't mention them on #320 (comment)

Sorry! I missed that bit. There my be a few files (a header in addition to a source file). The general rule is, inspect all changes to autogenerated (bootstrap) files and rollback those that are trivial (like a timestamp). But I see that in your case it is difficult to see if there are real changes or just those caused by a bison update. In this particular case, since you changed parser code in the .ypp file, it definitely needs regeneration (it wouldn't compile on the next non-incremental build with `RE2C_REBUILD_LEXERS=off`).

> I have a more recent bison version (3.5.1) that creates a huge diff. Is this acceptable?

Yes, I guess we'll have to live with this. If there are too many parser changes back and forth we can agree on some common bison version, but right now it's my system that's lagging behind and needs updating.

src/regexp/ast_to_re.cc
sergeyklay: I changed this one:

```diff
diff --git a/src/regexp/ast_to_re.cc b/src/regexp/ast_to_re.cc
index f0dd9cea..9017dbdb 100644
--- a/src/regexp/ast_to_re.cc
+++ b/src/regexp/ast_to_re.cc
@@ -60,7 +60,7 @@ static Range *char_to_range(RESpec &, const ASTChar &, bool);
 static Range *diff_to_range(RESpec &, const AST *);
 static Range *dot_to_range(RESpec &, const AST *);
 static Range *cls_to_range(RESpec &, const AST *);
-static bool misuse_of_named_def(RESpec &, const AST *);
+static void check_misuse_of_named_def(RESpec &, const AST *);
 static void assert_tags_used_once(RESpec &, const Rule &, const std::vector<Tag> &);
 static void init_rule(RESpec &, Rule &, const SemAct *, const std::vector<Tag> &, size_t, size_t);
 static bool is_icase(const opt_t *, bool);
@@ -252,10 +252,7 @@ RE *ast_to_re(RESpec &spec, const AST *ast, size_t &ncap, int32_t height)
             }

         case AST::REF:
-            if (misuse_of_named_def(spec, ast)) {
-                /* unreachable until 'misuse_of_named_def' uses 'exit(1)' */
-                return NULL;
-            }
+            check_misuse_of_named_def(spec, ast);
             return ast_to_re(spec, ast->ref.ast, ncap, height);

         case AST::ITER: {
@@ -368,11 +365,8 @@ Range *ast_to_range(RESpec &spec, const AST *ast)
             if (spec.opts->posix_syntax) break;
             return ast_to_range(spec, ast->cap);
         case AST::REF:
-            if (misuse_of_named_def(spec, ast)) {
-                /* unreachable until 'misuse_of_named_def' uses 'exit(1)' */
-                return NULL;
-            }
-          return ast_to_range(spec, ast->ref.ast);
+            check_misuse_of_named_def(spec, ast);
+            return ast_to_range(spec, ast->ref.ast);
         case AST::CLS:
             return cls_to_range(spec, ast);
         case AST::DOT:
@@ -444,7 +438,7 @@ RE *re_class(RESpec &spec, const loc_t &loc, const Range *r)
     return NULL; /* unreachable */
 }

-bool misuse_of_named_def(RESpec &spec, const AST *ast)
+void check_misuse_of_named_def(RESpec &spec, const AST *ast)
 {
     DASSERT(ast->type == AST::REF);

@@ -455,7 +449,6 @@ bool misuse_of_named_def(RESpec &spec, const AST *ast)
             , ast->ref.name->c_str());
         exit(1);
     }
-    return false;
 }

 void assert_tags_used_once(RESpec &spec, const Rule &rule
```

skvadrik (approved):
Great, thanks for taking care of all review comments!

PR #328, opened 2020-10-20
Restructuring Configure and Build jobs on GitHub Actions

sergeyklay:
This is a planned, intermediate change that prepares the following things:

- Preparing to use different generators (Xcode, NMake, Ninja)
- Preparing to use rebuild and test documentation (thanks to https://github.com/skvadrik/re2c/pull/326). This required renaming some steps
- Preparing to build libs
- Preparing to use Autotools
- Preparing to make a full migration of CMake builds from Travis CI

Notable improvements include the unification of the Configure job (now it is Fast Configure) for all platforms

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
           # Linux
           - name: linux-gcc-ubsan-ootree
             os: ubuntu-latest
-            # Undefined Behavior Sanitizer (cmake -DCMAKE_BUILD_TYPE=Ubsan)
+            generator: Unix Makefiles
             build-type: Ubsan
             build-dir: .build
             cxx-flags: '-O2'
@@ -28,7 +28,7 @@ jobs:
 
           - name: linux-gcc-asan-ootree
             os: ubuntu-latest
-            # Address Sanitizer (cmake -DCMAKE_BUILD_TYPE=Asan)
+            generator: Unix Makefiles
             build-type: Asan
             build-dir: .build
             cxx-flags: '-O2'
@@ -36,6 +36,7 @@ jobs:
 
           - name: linux-gcc-debug-ootree
             os: ubuntu-latest
+            generator: Unix Makefiles
             build-type: Debug
             build-dir: .build
             cxx-flags: -O2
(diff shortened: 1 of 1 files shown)


skvadrik (commented):
> Preparing to make a full migration of CMake builds from Travis CI

I'm not sure if you mean to remove Travis CI, or just to make it a subset of GitHub Actions. I think we should still keep it, for the following reasons: it's good to have some redundancy in case something happens to GitHub Actions, it is a different environment which may find different issues, it works and isn't broken, and it would take effort to setup once again if we remove it.

I have a few questions (maybe that's part of your plan which I don't fully understand, so can you explain a bit):

- Why is it necessary to rename `Configure` into `Fast Configure` and split it in two steps?

- Why is it necessary to rename `Build` step into `Fast Build`?

sergeyklay:
> I'm not sure if you mean to remove Travis CI,

I didn't plan to remove its support, at least in the near future, even though it is clearly slower than GitHub Action and provides less features. You are probably confused by the word "migration". Should I change the commit message?

> * Why is it necessary to rename `Configure` into `Fast Configure` 

Because there is another 'Configure' called 'Full Configure' with `-DRE2C_REBUILD_LEXERS`, `RE2C_BUILD_LIBS`, `RE2C_REBUILD_DOCS`. The same as Travis CI does right now:

Fast:
https://github.com/skvadrik/re2c/blob/8bfac019e5f3aecdacb9794421de35b67d51785b/.travis.yml#L47-L55

Full:
https://github.com/skvadrik/re2c/blob/8bfac019e5f3aecdacb9794421de35b67d51785b/.travis.yml#L58-L73

The same for 'Build' step. I don't mind any other name.

> and split it in two steps?

No, I just moved the preparation of the configure arguments, not configuration. The same as Travis CI does right now:
https://github.com/skvadrik/re2c/blob/8bfac019e5f3aecdacb9794421de35b67d51785b/.travis.yml#L29-L46


> Why is it necessary to rename Build step into Fast Build?
See above.

skvadrik:
> You are probably confused by the word "migration". Should I change the commit message?

No, it's absolutely fine, I just want to be on the same page.

skvadrik (approved):

PR #329, opened 2020-10-20
Complete CI implementation using CMake and GitHub  Actions

sergeyklay:
This PR adds ability to build re2c on GitHub Actions with the following options:

- **`-DRE2C_REBUILD_LEXERS=ON`**: Windows, macOs, Linux
- **`-DRE2C_BUILD_LIBS=ON`**: Windows, macOs, Linux
- **`-DRE2C_REBUILD_DOCS=ON`**: macOs, Linux

With this PR I completely implement CI on GitHub Actions using CMake. Current workflow configuration fully cover the Travis' configuration (even more).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,6 +159,26 @@ jobs:
               ;;
           esac
 
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.x
+          architecture: x64
+
+      - name: Get PIP Cache Key
+        id: pip-cache
+        run: |
+          python -c "from pip._internal.locations import USER_CACHE_DIR; \
+                     print('::set-output name=dir::' + USER_CACHE_DIR)"
+
+      - name: Install Python Dependencies
+        # Disabled until this https://github.com/skvadrik/re2c/issues/327
+        # will be implemented
+        if: startsWith(runner.os, 'Windows') == false
+        run: |
+          python -m pip install --upgrade pip
+          pip install docutils pygments
+
       - name: Prepare Configure Command
         id: configure-command
         run: |
@@ -179,7 +199,6 @@ jobs:
 
           echo "::set-output name=cmake::${args[@]}"
 
-      #
       - name: Fast Configure
         run: |
           # Read prepared command into an array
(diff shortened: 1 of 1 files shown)


skvadrik (approved):
Nice! I have a few questions/suggestions for the future:
- We can merge skeleton configurations with normal configurations and run skeleton tests after normal tests. They should be enabled for a few selected configurations (such as linux-gcc-debug-ootree), not for every one. It's only +2 minutes, the total time will still be less than a sanitizer build.
- Why is it a problem to run tests on Windows (in Cygwin environment)?

skvadrik:
Thanks and sorry for the delayed review!

Please don't feel obliged to wait for my review if that's an obstacle for you. I'm glad you do, but I shouldn't be holding back your work. If something is wrong it can always be fixed in a follow-up commit.

sergeyklay:
> We can merge skeleton configurations with normal configurations and run skeleton tests after normal tests. They should be enabled for a few selected configurations (such as linux-gcc-debug-ootree), not for every one. It's only +2 minutes, the total time will still be less than a sanitizer build.

Just to clarify: Do you suggest enable skeleton test for for some more configurations and merge with 'Standard Tests' job?
 If so, could you please list the configurations in which we should include skeleton test?

> * Why is it a problem to run tests on Windows (in Cygwin environment)?

As I said before, I'd like to write pure PowerShell wrapper for this. This is in my plans. As for Cygwin: I'd like to provide a separated Cygwin matrix item to go all the way, from build to tests. This will be done on purpose so as not to mix Cygwin with Visual Studio.

skvadrik:
> Just to clarify: Do you suggest enable skeleton test for some more configurations and merge with 'Standard Tests' job?
If so, could you please list the configurations in which we should include skeleton test?

Yes, I had something like this in mind: https://github.com/skvadrik/re2c/pull/330.

> As I said before, I'd like to write pure PowerShell wrapper for this. This is in my plans. As for Cygwin: I'd like to provide a separated Cygwin matrix item to go all the way, from build to tests. This is planned on purpose to mix Cygwin with Visual Studio. 

Cool!

> You may have noticed that the current Windows builds proves that the Windows users can build and even test re2c using native tools, without resorting to Mingw, Cygwin and so on.

Yes, that's nice.

PR #330, opened 2020-10-20
GitHub Actions: run skeleton tests after normal tests in the same build.

skvadrik:


.github/workflows/ci.yml
```diff
@@ -64,18 +56,10 @@ jobs:
             build-type: Release
             build-dir: .build
             cxx-flags: ${{ null }}
-            skeleton: ${{ false }}
+            skeleton: ${{ true }}
```
sergeyklay, line 59: Here and in all other places: could you please rename `linux-gcc-release-ootree` to  `linux-gcc-release-ootree-skeleton` to follow common naming pattern?

skvadrik: Done!

sergeyklay (approved):
I'm not sure about `"${{ matrix.skeleton }}" == true`. However, if I'm wrong and it actually works as you expect - great!

sergeyklay:
Just for your info: There is an ability to cancel builds in case of force-push

image

skvadrik:
> Just for your info: There is an ability to cancel builds in case of force-push

Ah, I was wondering about that. I hoped it would auto-cancel, but apparently not (or maybe they get cancelled, but still show in the UI). :)

skvadrik:
> I'm not sure about `"${{ matrix.skeleton }}" == true`

Seems to work as expected (skeleton build runs skeleton tests, non-skeleton one doesn't). Are you worries about the usual shell things like empty variable on the left hand side, or not quoting `true`? Maybe it should be `"x${{ matrix.skeleton }}" == "xtrue"`.

sergeyklay:
> Maybe it should be `"x${{ matrix.skeleton }}" == "xtrue"`.

Yeah, I usually prefer something like this. But, if everything is OK, then great!

.github/workflows/ci.yml
```diff
@@ -275,11 +259,14 @@ jobs:
...
-
-      - name: Skeleton Validation
-        if: startsWith(runner.os, 'Windows') == false && matrix.skeleton == true
+      - name: Test
+        if: startsWith(runner.os, 'Windows') == false
         working-directory: ${{ matrix.build-dir }}
-        run: ./run_tests.sh --skeleton
+        run: |
+          # main test suite
+          cmake --build . --target check
```
sergeyklay, line 267: It should be:

```suggestion
          cmake --build ${{ matrix.build-dir }} --target check
```

skvadrik: But I have added `working-directory: ${{ matrix.build-dir }}`, doesn't it mean that `.` is `${{ matrix.build-dir }}` ?

sergeyklay: Oh! You're right!

skvadrik: Also, the tests do run.

PR #332, opened 2020-10-31
Use CMAKE_BUILD_TYPE for single-config build types only

sergeyklay:
Hello,

There are some commits I'd like to separate from my current working branch and push them to master.  They are enough trivial and I hope provided patch doesn't need additional description:

- Use `CMAKE_BUILD_TYPE` for single-config build types only
- Ignore autogenerated directories and files
- Add `.log` extension for auto generated test logs


.gitignore
```diff
@@ -1,4 +1,19 @@
...
+# https://help.github.com/articles/ignoring-files/#create-a-global-gitignore
+
+# The directories to ignore (do not add trailing '/'s, they skip symlinks).
+/autom4te.cache
+/m4
+/test_*
+/.build*
+/build
+
+# The files auto generated by automake.
```
skvadrik, line 16: "autogenerated" without a space I think

.gitignore
```diff
@@ -1,4 +1,19 @@
...
+# for ignoring files in every Git repository on your computer.
+#
+# https://help.github.com/articles/ignoring-files/#create-a-global-gitignore
+
+# The directories to ignore (do not add trailing '/'s, they skip symlinks).
+/autom4te.cache
+/m4
+/test_*
+/.build*
+/build
```
skvadrik, line 14: Why these two? `/build` is a subdirectory that contains scripts (and shouldn't be ignored), and `/.build` is used by some scripts as the name of the build directory, but it's configurable.

.github/workflows/ci.yml
```diff
@@ -167,14 +167,20 @@ jobs:
         id: configure-command
         run: |
           # Common CMake arguments
-          args=( -S . -B ${{ matrix.build-dir }} -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} )
+          args=( -S . -B ${{ matrix.build-dir }} )
 
           # Generator's specific stuff
           # We don't add generator here to avoid splitting it before
```
skvadrik, line 173: Not related to this change, but I realized that I don't understand what this comment means. Can you clarify it?

skvadrik (approved):

skvadrik (changes requested):
Nice cleanup. I wanted to approve one of commits, as I have a few questions about ignored files.

.gitignore
sergeyklay: We use `/.build`, `/.build_*` and so on in our scripts. I'll remove `/build`

.github/workflows/ci.yml
sergeyklay: Consider the following example:
```bash
args=(-DCMAKE_BUILD_TYPE=Debug -G "Unix Makefiles")

str_args=$(echo "${args[@]}")
arr_args=($str_args)

for i in "${arr_args[@]}"; do echo "ARG: $i"; done
```

Expected output:
```
ARG: -DCMAKE_BUILD_TYPE=Debug
ARG: -G
ARG: Unix Makefiles
```

Actual output:
```
ARG: -DCMAKE_BUILD_TYPE=Debug
ARG: -G
ARG: Unix
ARG: Makefiles
```

This is why we don't add generator in 'Prepare Configure Command' step

skvadrik: Ah, I see, thanks! I couldn't understand that it refers to word split. Can you reword the comment as "We don't add generator here to avoid word split when expanding array elements, e.g. `Unix Makefiles` passed as two arguments".

skvadrik: (Not necessarily in this PR.)

skvadrik (approved):

.github/workflows/ci.yml
sergeyklay: Could you please take a look https://github.com/skvadrik/re2c/pull/332/commits/5e7ac621cdfa72f65ceea75e38b7ac711fff9097

skvadrik (approved):

PR #336, opened 2020-11-07
Speed up the CMake configuration by avoiding unnecessary checks

sergeyklay:


skvadrik (commented):
A few minor comments. Thanks for the patch!

skvadrik (approved):

PR #338, opened 2020-12-29
Fix compiling on Windows (MSVC)

unixod:
Hi,

By trying to build re2c on Windows, using MSVS 2017 I faced the compile-time error: C2039: 'isalnum': is not a member of 'std'. This PR fixes this issue.

By the way I added a small fix of using this function: in order to avoid UB (which may occur in some situations, see links below) the argument should be passed by preliminary converting it to unsigned char. More info about this issue:
-  https://stackoverflow.com/a/39520176/1117306
-  https://en.cppreference.com/w/cpp/string/byte/isalnum.

### DIFF
--- a/src/compile.cc
+++ b/src/compile.cc
@@ -4,6 +4,7 @@
 #include <ostream>
 #include <string>
 #include <vector>
+#include <cctype>
 
 #include "src/adfa/adfa.h"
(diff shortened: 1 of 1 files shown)


skvadrik:
Thank you!

PR #339, opened 2020-12-29
Enable support of C++ exceptions on MSVC builds.

unixod:
HI,

By looking at build log on Windows (MSVC) I noticed a number of warnings of the form:
`'noexcept' used with no exception handling mode specified; termination on exception is not guaranteed. Specify /EHsc (compiling source file ...src\codegen\gen_state.cc)`

If I recall correctly from my experience, if this flag isn't set then we may encounter situations when destructors of automatic objects aren't called during the exception propagation process. 

The following doc also recommends (if project doesn't use SEH) to set this flag:
https://docs.microsoft.com/ru-ru/cpp/build/reference/eh-exception-handling-model?view=msvc-160

### DIFF
--- a/cmake/Re2cCompilerFlags.cmake
+++ b/cmake/Re2cCompilerFlags.cmake
@@ -56,6 +56,7 @@ else()
     # /Wall enables all warnings that are off by default including
     # some of the ones GCC/Clang enables through -Wxxxx flags.
     try_cxxflag("/Wall")
+    try_cxxflag("/EHsc")    # Support C++ exceptions.
 endif()
 
 # Print compiler and linker flags


skvadrik:
@unixod Thanks! I wonder which version of MSVC you are using, as I see no such warnings in the CI logs: https://github.com/skvadrik/re2c/runs/1622655634, which uses Visual Studio 2019 (`C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\HostX64\x64\CL.exe`).

unixod:
> @unixod Thanks! I wonder which version of MSVC you are using, as I see no such warnings in the CI logs: https://github.com/skvadrik/re2c/runs/1622655634, which uses Visual Studio 2019 (`C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\HostX64\x64\CL.exe`).

Ah, I see. Now I realized that disappearance of /EHsc in my case was caused by using -DCMAKE_CXX_FLAGS=/MP in command line on our CI script (i.e. default flags were inadvertently overwritten). Be that as it may, I think it is good choice to enforce  using of this flag regardless of what users put in CMAKE_CXX_FLAGS because they may accedently, as it happend in my case, disable this important feature)

skvadrik:
Ok, let's keep `/EHsc`.

I noticed that in Visual Studio builds whatever we set in https://github.com/skvadrik/re2c/blob/master/cmake/Re2cCompilerFlags.cmake gets overridden by `CMAKE_CXX_FLAGS` that is populated by CMake (it ends up being `/EHsc /Wall /DWIN32 /D_WINDOWS /W3 /GR /EHsc`, where `/W3` overrides `/Wall`). This happens because we _prepend_ flags to this variable instead of _appending_ them (to avoid overriding the flags set by the user on the command-line). Usually it works because CMake uses build-type specific variables `CMAKE_CXX_FLAGS_DEBUG`, `CMAKE_CXX_FLAGS_RELEASE` etc. to set defaults instead of `CMAKE_CXX_FLAGS`, but this is not the case for multiconfiguration builds used by Visual Studio.

I fixed this by changing the logic to override `CMAKE_CXX_FLAGS` in MSVC builds (https://github.com/skvadrik/re2c/commit/ee6e6d7eb8d0912eaa6e6491cbdd1cbf8c26807c). Now the flags come in this order: `/DWIN32 /D_WINDOWS /W3 /GR /EHsc  /Wall /EHsc`, and there is a flood of warnings.

unixod:
Thanks!

PR #340, opened 2021-01-04
Benchmarks

skvadrik:


.github/workflows/ci.yml
```diff
@@ -58,6 +58,14 @@ jobs:
             cxx-flags: ${{ null }}
             skeleton: ${{ true }}
 
+          - name: linux-gcc-benchmark-ootree
+            os: ubuntu-latest
+            generator: Unix Makefiles
+            build-type: Benchmark
```
sergeyklay, line 64: Is there a new build type? Did'n noticed the relevant changes in the CMake Lists

skvadrik: Yes, I was working on a branch `benchmarks`, here is the change: https://github.com/skvadrik/re2c/commit/98dfdcaf651cd2e96892f37f29ad02e8f429f2a4

The new build type takes a long time on my local machine, so I'm trying to see how long it will take on Github Actions.

skvadrik: TBH I started this pull request to trigger CI checks, as the new build seems to be stuck on and I was unsure why. It's probably not stuck, but just takes a long time, because when I cancel it I'm able to see the partial build log (and it builds fine without errors). If it takes too long, I can pick a subset of small benchmarks for CI.

sergeyklay: >  I'm trying to see how long it will take on Github Actions.

AFIK, GitHub runners ships with 2 CPU cores (plus HT), so I don't believe it will be faster (if your PC is not Raspberry PI)

sergeyklay: Screenshot 2021-01-04 at 15 04 14

skvadrik: Good to know, thanks. Now that the jobs have finished I can see that it takes ~30 min, which is too long. I'll close this PR for now and add an option to build a fast subset of benchmarks (most of them are ok, but there are a few files that take about 5min to compile with GCC/Clang because the generated automata are large).

sergeyklay: @skvadrik Wait. Let's configure workflow to run bechmarks via cron on regular basis. Let's say Sunday nights? We can make the process fully automatic, as well as send a report by mail. 

> it takes ~30 min

AFAIK, GitHub imposes a time limit on the task execution time. I do not remember the exact time now, but it was not bad to have it in mind

sergeyklay: The maximum number of minutes to let a workflow run before GitHub automatically cancels it is 360:

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

skvadrik: > Let's configure workflow to run bechmarks via cron on regular basis. Let's say Sunday nights? We can make the process fully automatic, as well as send a report by mail.

Great idea! Then I can actually run them in addition to making sure that they build. Running would take much more time, as there are very slow cases that show disadvantages of some algorithms. Do we need a separate workflow for this?

sergeyklay: Yes, I think we do. Lets create a simple workflow with something like this in the header:

```yml
on:
  schedule:
    - cron: '0 0 * * 7'
```

skvadrik: Ok, I'll experiment with this on `benchmarks` branch.

skvadrik: The new workflow for benchmarks is here: https://github.com/skvadrik/re2c/commit/2b3fb5a135262e502b92a2dbf45b7b05e40514df, example run: https://github.com/skvadrik/re2c/actions/runs/464845130 (it should run weekly or when triggered manually, but that only works on default branch, so I used push event for testing). @sergeyklay Can you please take a look?

sergeyklay: Looks pretty neat to me! Let's merge it :)

PR #348, opened 2021-03-31
configure.ac: avoid bash-specific '==' comparison

trofi:
In https://bugs.gentoo.org/779187 Matt noticed errors when running ./configure
on dash:
    ./configure: 5339: test: x: unexpected operator
    ./configure: 5386: test: x: unexpected operator

The change is to use POSIX '=' comparison instead of bash-specific ==.

Patch-by: Matt Whitlock
Bug: https://bugs.gentoo.org/779187


skvadrik:
Thank you @trofi !

PR #359, opened 2021-06-16
Run regular re2c tests on macOs too

sergeyklay:
I noticed that we don't run regular re2c tests for macOs. `macos-clang-debug-ootree-skeleton` and `macos-clang-debug-intree-skeleton` are very similar, so I changed `skeleton: ${{ true }}` to `skeleton: ${{ false }}` to run non-skeleton tests on macOs too.

With change we'll have the following tests matrix:

**Linux**

- skeleton tests: 3
- non-skeleton tests: 2

**macOS**

- skeleton tests: 2
- non-skeleton tests: 1  (I introduced this)

**Windows**

- skeleton tests: 0
- non-skeleton tests: 3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,13 +59,13 @@ jobs:
             skeleton: ${{ true }}
 
           # macOS
-          - name: macos-clang-debug-ootree-skeleton
+          - name: macos-clang-debug-ootree
             os: macos-latest
             generator: Unix Makefiles
             build-type: Debug
             build-dir: .build
             cxx-flags: -O2
-            skeleton: ${{ true }}
+            skeleton: ${{ false }}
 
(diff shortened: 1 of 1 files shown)


skvadrik:
The naming may be confusing, but "skeleton" configurations run regular tests as well, they just run skeleton tests in addition. Skeleton tests do not require any special build type, it is just an option of run_tests.sh. For example, if you expand build https://github.com/skvadrik/re2c/pull/359/checks?check_run_id=2836412286 (the "Test" step), you can see that it runs regular tests before the skeleton ones (it is easy to distinguish them visually because skeleton tests have many `(expected re2c error)` messages after `OK` and the name of the test).

sergeyklay:
Oh! My bad. I completely missed this

PR #367, opened 2021-06-23
configure.ac: allow user to set non-standard PYTHON

trofi:
Useful on systems with multiple python implementations.
While at it allow both 'python3' (preferred) and 'python' spellings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,7 @@ option(RE2C_REGEN_BENCHMARKS "Regenerate C code for benchmarks" OFF)
 
 # checks for programs
 find_package(BISON)
+find_package(Python3 COMPONENTS Interpreter)
 
 # checks for C++ compiler flags
 set(CMAKE_CXX_STANDARD 98)
@@ -96,11 +97,11 @@ set(re2c_docs
 
 set(top_srcdir "${CMAKE_CURRENT_SOURCE_DIR}")
 set(top_builddir "${CMAKE_CURRENT_BINARY_DIR}")
+set(PYTHON "${Python3_EXECUTABLE}")
 
 configure_file(doc/manpage.rst.in doc/manpage.rst @ONLY)
 configure_file(doc/help.rst.in doc/help.rst @ONLY)
 
-find_package(Python3 REQUIRED COMPONENTS Interpreter)
 configure_file(run_tests.py.in run_tests.py @ONLY)
 set(RE2C_RUN_TESTS "${CMAKE_CURRENT_BINARY_DIR}/run_tests.py")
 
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -85,6 +85,7 @@ AM_CONDITIONAL([REGEN_BENCHMARKS],
 
 # checks for programs
 AC_PATH_PROG(BISON, bison, no)
+AC_CHECK_PROGS(PYTHON, [python3 python])
```
skvadrik, line 88: Note that python2 shall not pass!

CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ option(RE2C_REGEN_BENCHMARKS "Regenerate C code for benchmarks" OFF)
 
 # checks for programs
 find_package(BISON)
+find_package(Python)
```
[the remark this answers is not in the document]
sergeyklay: I think we can provide an ability to pass PYTHON from cmake command line, i.e. 
```
-DPYTHON=/usr/local/bin/python3.10
```
and use 
```
find_package(Python3 REQUIRED COMPONENTS Interpreter)
```
 if the path was not explicitly specified.

Also I think we can introduce something like `WITH_TESTS` to not search Python at all if the user doesn't need tests

configure.ac
trofi: My assumption here is that systems might have python3 under two names in system: `python3` (probably most popular) and `python` (for ones that removed `python2` and don't plan to go back).

CMakeLists.txt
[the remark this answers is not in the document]
trofi: Moved `find_package(Python3 REQUIRED COMPONENTS Interpreter)` above and reused. While at it dropped `REQUIRED`. Did not do anything more drastic.

configure.ac
skvadrik: Right, makes sense.

CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ option(RE2C_REGEN_BENCHMARKS "Regenerate C code for benchmarks" OFF)
 
 # checks for programs
 find_package(BISON)
+find_package(Python3 COMPONENTS Interpreter)
```
skvadrik, line 51: @sergeyklay To clarify, for now we do not require python and do not bother with the checks (if it was found) either.

sergeyklay (approved):

trofi:
For posterity: looks like cmake way to set non-standard python is via `Python3_EXECUTABLE`: `cmake .. -DPython3_EXECUTABLE=/usr/bin/python3.10`.

skvadrik (approved):

skvadrik:
Thank you @trofi !

PR #374, opened 2021-08-11
Provide support of CMake Presets

sergeyklay:
Since 3.19.8 CMake looks in the top source directory of a project for files named `CMakePresets.json` and `CMakeUserPresets.json`. Neither file is required to be present, but if the user tries to list or use presets, at least one of the files must exist. If both files are present, they will effectively be merged by reading the `CMakePresets.json` file first and then the `CMakeUserPresets.json` file.

`CMakePresets.json` can be used to define presets for things like continuous integration jobs or cross-compilation setups using toolchains provided as part of the project. In general this feature allows us to simplify CI configuration phase by taking out all the logic into presets-file and helps us to make reproducible builds. From now on, you no longer need to carefully read and copy all the flags locally in order to reproduce CI configuration or build. Simple use the same one line command as at CI.

As a bonus, users of modern CMake can use a cross-platform and portable way instead of `build/__cmake*.sh` scripts.

To see the list of available presets use:

    cmake --list-presets

To configure re2c use:

    cmake --preset=<NAME>

To build re2c use:

    cmake --build --preset=<NAME>

To test re2c use:

    cmake --build --preset=<NAME> --target tests

References:

- https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html
- https://gitlab.kitware.com/cmake/cmake/-/issues/21164
- https://cmake.org/cmake/help/latest/guide/user-interaction/index.html#presets


.github/workflows/ci.yml
```diff
@@ -158,108 +101,44 @@ jobs:
...
-        working-directory: ${{ matrix.build-dir }}
+      - name: Run Main Test Suite
+        run: cmake --build --preset=${{ matrix.name }}-full --target tests
+
+      - name: Get build dir
+        id: build-info
         run: |
-          # main test suite
-          cmake --build . --target check
+          build_dir="$(cmake --build --preset=${{ matrix.name }}-full --target print_bin_dir | tail -n 2 | head -n 1)"
```
skvadrik, line 134: The head-tail command needs a comment to make it easier to understand.

sergeyklay: Bad news:

File version must be 3 or higher for condition support like this:

```json
      "condition": {
        "type": "equals",
        "lhs": "${hostSystemName}",
        "rhs": "Linux"
      }
```

Version 3 was introduced in CMake 3.21.1. 

Also `installDir` is allowed in preset files specifying version 3 or above.

CMakePresets.json
```diff
@@ -0,0 +1,592 @@
...
+      "inherits": "linux-gcc-ubsan-ootree-base",
+      "installDir": "${sourceDir}/install-2",
+      "cacheVariables": {
+        "RE2C_BUILD_RE2GO": "ON",
+        "RE2C_BUILD_LIBS": "ON",
+        "RE2C_REBUILD_LEXERS": "ON",
+        "RE2C_REBUILD_DOCS": "ON",
+        "RE2C_FOR_BUILD": "${sourceDir}/install/bin/re2c"
+      }
+    },
```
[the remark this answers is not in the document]
sergeyklay: Let me think about it a bit, maybe I can cut it down. In any case, you should treat this as a declarative configuration without conditions support

CMakePresets.json
```diff
@@ -0,0 +1,592 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 20,
+    "patch": 0
```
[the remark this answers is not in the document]
sergeyklay: Bad news:

File version must be 3 or higher for condition support like this:

```json
      "condition": {
        "type": "equals",
        "lhs": "${hostSystemName}",
        "rhs": "Linux"
      }
```

Version 3 was introduced in CMake 3.21.1. 

Also `installDir` is allowed in preset files specifying version 3 or above.

CMakeLists.txt
```diff
@@ -530,3 +530,5 @@ if (CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
     add_custom_target(check)
     add_dependencies(check check_re2c check_libre2c)
 endif()
+
+add_custom_target(print_bin_dir ${CMAKE_COMMAND} -E echo "${CMAKE_BINARY_DIR}")
```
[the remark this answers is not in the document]
sergeyklay: I'm still not sure if this should be merged. Let me find the best solution

.github/workflows/ci.yml
sergeyklay: See https://github.com/skvadrik/re2c/pull/374#discussion_r687241746

CMakePresets.json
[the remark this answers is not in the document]
skvadrik: Ah ok, I was just curious why that particular version. It doesn't make much difference since it's not imposing the restriction on the main CMakeLists.txt.

.github/workflows/ci.yml
skvadrik: Ok, I'll mark this as resolved then.

sergeyklay:
@skvadrik I managed to reduce the total number of presets through inheritance. Take a look at this commit https://github.com/skvadrik/re2c/pull/374/commits/4ee6ea6cc5ea4d79330886842abfecfaa1cc404a I deleted 11 presets 

Also, I'd like to note that we have a similar approach in current workflow.

Base config:
https://github.com/skvadrik/re2c/blob/aa2da70299b7807391a8a27b7e1ab3e9f494a205/.github/workflows/ci.yml#L161-L188

Fast config:
https://github.com/skvadrik/re2c/blob/aa2da70299b7807391a8a27b7e1ab3e9f494a205/.github/workflows/ci.yml#L190-L206

Full config:
https://github.com/skvadrik/re2c/blob/aa2da70299b7807391a8a27b7e1ab3e9f494a205/.github/workflows/ci.yml#L220-L239

The same for build.

CMakePresets.json
```diff
@@ -0,0 +1,528 @@
...
+        "RE2C_REBUILD_DOCS": false,
+        "RE2C_FOR_BUILD": "${sourceDir}/install/bin/re2c"
+      }
+    }
+  ],
+  "buildPresets": [
+    {
+      "name": "linux-build-base",
+      "hidden": true,
+      "configuration": "Debug",
```
skvadrik, line 354: Why do we have a default here? Isn't it better to explicitly specify in each preset, to make them symmetric? Same question for macOS and Windows, and for configurations.

CMakePresets.json
```diff
@@ -0,0 +1,528 @@
...
+      "condition": {
+        "type": "equals",
+        "lhs": "${hostSystemName}",
+        "rhs": "Linux"
+      }
+    },
+    {
+      "name": "linux-gcc-ubsan-ootree-fast",
+      "displayName": "Linux x64: GCC / Ubsan / Out of tree (fast)",
```
skvadrik, line 363: Maybe x86_64 (here and in other places)?

CMakePresets.json
```diff
@@ -0,0 +1,528 @@
...
+    {
+      "name": "linux-gcc-asan-ootree-full",
+      "displayName": "Linux x64: GCC / Asan / Out of tree (full)",
+      "configurePreset": "linux-gcc-asan-ootree-full",
+      "inherits": "linux-build-base",
+      "configuration": "Asan"
+    },
+    {
+      "name": "linux-gcc-debug-ootree-skeleton-fast",
+      "displayName": "Linux x64: GCC / Skeleton / Out of tree (fast)",
```
skvadrik, line 391: Add "Debug" in the `displayName`? Likewise for other presets below (and above, in configurations). Asan/Ubsan build types are fine as is, they are special.

CMakePresets.json
```diff
@@ -0,0 +1,528 @@
...
+    },
+    {
+      "name": "linux-gcc-debug-intree-m32-skeleton-full",
+      "displayName": "Linux x86: GCC / Skeleton / In tree (full)",
+      "configurePreset": "linux-gcc-debug-intree-m32-skeleton-full",
+      "inherits": "linux-build-base"
+    },
+    {
+      "name": "linux-gcc-release-ootree-skeleton-fast",
+      "displayName": "Linux x64: GCC / Release / Out of tree (fast)",
```
skvadrik, line 415: Release presets are missing "Skeleton" in the `displayName` (configurations above as well).

CMakePresets.json
```diff
@@ -0,0 +1,528 @@
...
+      "displayName": "Windows x64: MSVC / Debug / Out of tree (fast)",
+      "configurePreset": "windows-msvc-debug-ootree-fast",
+      "inherits": "windows-build-base"
+    },
+    {
+      "name": "windows-msvc-debug-ootree-full",
+      "displayName": "Windows x64: MSVC / Debug / Out of tree (full)",
+      "configurePreset": "windows-msvc-debug-ootree-full",
+      "inherits": "windows-build-base"
+    },
```
skvadrik, line 500: I think we can drop this pair (in a separate commit). One debug/release and one intree/ootree is enough. Likewise for macOS and for configuration presets.

CMakePresets.json
```diff
@@ -0,0 +1,528 @@
...
+      "displayName": "Linux x64: GCC / Ubsan / Out of tree (full)",
+      "inherits": "linux-gcc-ubsan-ootree-fast",
+      "installDir": "${sourceDir}/install-2",
+      "cacheVariables": {
+        "RE2C_BUILD_RE2GO": true,
+        "RE2C_BUILD_LIBS": true,
+        "RE2C_REBUILD_LEXERS": true,
+        "RE2C_REBUILD_DOCS": true,
+        "RE2C_FOR_BUILD": "${sourceDir}/install/bin/re2c"
+      }
```
skvadrik, line 51: Can this block be inherited? As far as I can tell it is duplicated many times (in all "full" presets).

sergeyklay: Well, more recently, you hinted that there are too many presets :) Of course, we can do more atomic presets and even use multiple inheritance.

CMakePresets.json
sergeyklay: I don't seem to fully understand how you want to do this since `configurePreset` specifying the name of a configure preset to associate with this build preset. As documentation says:

> If `configurePreset` is not specified, it **must** be inherited from the inherits preset (unless this preset is hidden).

It seems to me we can't use one build preset for two different config presets. But lets investigate this a bit later.

CMakePresets.json
sergeyklay: Please take a look: https://github.com/skvadrik/re2c/pull/374/commits/216932105275c3017cf0bf8bbb05c2492d09fa3d

CMakePresets.json
sergeyklay: Please take a look: https://github.com/skvadrik/re2c/pull/374/commits/216932105275c3017cf0bf8bbb05c2492d09fa3d

CMakePresets.json
sergeyklay: Please take a look: https://github.com/skvadrik/re2c/pull/374/commits/216932105275c3017cf0bf8bbb05c2492d09fa3d

CMakePresets.json
sergeyklay: If I understood you correctly, then here is my patch: https://github.com/skvadrik/re2c/pull/374/commits/2714a357b0b24ea430e53cb2c53f631d48c51aed

skvadrik: What you did is exactly what I meant, thanks.

CMakePresets.json
skvadrik: No, I mean a very simple thing, not related to the presets and your PR. I noticed that we have the following combinations for macOS/Windows:
- debug ootree
- debug intree
- release ootree

And it seems to me that we should have only:
- debug intree
- release ootree

CMakePresets.json
skvadrik: Your change is exactly what I meant --- I know it's a few more lines of code, but I'm willing to trade that off for symmetry in the presets.

skvadrik (approved):
LGTM now. The remaining comment about reducing a few combinations for macOS/Windows is not necessary to include in this PR. Maybe later.

skvadrik (approved):

CMakePresets.json
sergeyklay: A, got it!

PR #376, opened 2021-09-03
Define _POSIX_SOURCE for fdopen in cygwin

Biswa96:
This fixes the error:
error: ‘fdopen’ was not declared in this scope; did you mean ‘fopen’?


skvadrik:
Thanks for the patch!

PR #378, opened 2021-11-02
Setup Java for the benchmarks workflow using actions/setup-java

nightlark:
Use `actions/setup-java` instead of `actions/setup-python` for the `Setup Java` step in the benchmarks workflow. Something seemed off about the action that was currently being used for setting up Java.


skvadrik:
Thanks! I think it was copy-paste.

PR #381, opened 2021-11-09
Initial implementation of --loop-switch for Rust support

Raekye:
With this option, instead of the usual goto/label approach, the DFA is wrapped in a loop with a `switch` statement on a `yystate` variable. Each DFA state is an individual `case`. Transitions between states are done by setting the `yystate` variable to the index of the target state and `continue`ing the loop.

This is option is necessary in order to support Rust as a future target language as Rust does not have `goto`s. See the commit message for further details.


skvadrik:
Thank you \o/

PR #382, opened 2021-11-17
fix struct/class declaration mismatch (-Wmismatched-tags)

trofi:
Detected by clang's-Weverything:

    ./src/regexp/re.h:17:1: warning: struct 'Msg' was previously declared as a class;
      this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
    struct Msg;
    ^


skvadrik:
\o/

PR #383, opened 2021-11-17
avoid using reserved identifiers (underscores)

trofi:
clang with -Weverything detects these as:

    ./src/util/free_list.h:9:16:
      warning: identifier '_Ty' is reserved
      because it starts with '_' followed by a capital letter [-Wreserved-identifier]
    template<class _Ty>
                   ^


skvadrik:
Thank you @trofi !

PR #387, opened 2021-11-27
CMakePresets.json: inherit "full" configuration before "fast".

skvadrik:
Otherwise variable definitions from "fast" configuration take precedence
over the "full" configuration, resulting in disabling some important
configrations, such as Go language support.

For example, if "fast" is inherited before "full" we have the following
incorrect configrations that sets RE2C_BUILD_RE2GO:BOOL="FALSE":

  $ cmake -S .. --preset=linux-gcc-release-ootree-skeleton-full
  Preset CMake variables:

    CMAKE_BUILD_TYPE="Release"
    CMAKE_CXX_COMPILER="g++"
    CMAKE_C_COMPILER="gcc"
    CMAKE_INSTALL_PREFIX:PATH="[CUT]/re2c/install"
    RE2C_BUILD_LIBS:BOOL="TRUE"
    RE2C_BUILD_RE2GO:BOOL="FALSE"
    RE2C_FOR_BUILD="[CUT]/re2c/install/bin/re2c"
    RE2C_REBUILD_DOCS:BOOL="TRUE"
    RE2C_REBUILD_LEXERS:BOOL="TRUE"

But if "full" is inherited after "fast" we have the following (correct):

  $ cmake -S .. --preset=linux-gcc-release-ootree-skeleton-full
  Preset CMake variables:

    CMAKE_BUILD_TYPE="Release"
    CMAKE_CXX_COMPILER="g++"
    CMAKE_C_COMPILER="gcc"
    CMAKE_INSTALL_PREFIX:PATH="[CUT]/re2c/install-2"
    RE2C_BUILD_LIBS:BOOL="TRUE"
    RE2C_BUILD_RE2GO:BOOL="TRUE"
    RE2C_FOR_BUILD="[CUT]/re2c/install/bin/re2c"
    RE2C_REBUILD_DOCS:BOOL="TRUE"
    RE2C_REBUILD_LEXERS:BOOL="TRUE"

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -53,7 +53,7 @@
     {
       "name": "linux-gcc-ubsan-ootree-full",
       "displayName": "Linux x86_64: GCC / Ubsan / Out of tree (full)",
-      "inherits": ["linux-gcc-ubsan-ootree-fast", "default-full-configuration"]
+      "inherits": ["default-full-configuration", "linux-gcc-ubsan-ootree-fast"]
     },
     {
       "name": "linux-gcc-asan-ootree-fast",
@@ -71,7 +71,7 @@
     {
       "name": "linux-gcc-asan-ootree-full",
       "displayName": "Linux x86_64: GCC / Asan / Out of tree (full)",
-      "inherits": ["linux-gcc-asan-ootree-fast", "default-full-configuration"]
+      "inherits": ["default-full-configuration", "linux-gcc-asan-ootree-fast"]
     },
     {
       "name": "linux-gcc-debug-ootree-skeleton-fast",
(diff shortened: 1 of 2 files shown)


skvadrik:
@sergeyklay https://github.com/skvadrik/re2c/pull/387/commits/2017418cd65d28eaed12713c6c96b06cd851674c could be interesting to you as it was not at all obvious that we were testing a slightly wrong configuration (I noticed it accidentally). See one of the old builds that disabled Go support in the "full" config: https://github.com/skvadrik/re2c/runs/4293511127?check_suite_focus=true#step:12:14

sergeyklay:
Very interesting 🤔 Thank you!

PR #396, opened 2022-02-20
run_tests.py.in: add --verbose flag, disable successful printing by d…

trofi:
…efault

re2c test are so fast that text output in terminal takes comparable time
on slower terminals. Let's disable it by default. Can still be brought
back with --verbose option.


skvadrik:
Thanks! \o/

skvadrik:
The failed CI builds are due to the recent Github migration to the new MSVC toolchain (unrelated to this PR).

PR #400, opened 2022-03-30
.github/workflows/ci.yml: set up trivial apt cache

trofi:
CI runs occasionally fail to fetch apt packages due to unavailable
remote. Let's try to cache heavyweight .deb packages and see if it
helps.

.deb files should not normally change frequently.


skvadrik:
Thank you \o/

PR #415, opened 2022-09-03
bison parsers: switch to pure bison API

trofi:
Before the change parsers relied on `YYSTYPE` global variable.
Not very handy for libre2c if it's expected to be used in reentrant
or multi-threaded contexts.

The change enables `%define api.pure full` and adjusts lexer calls
to explicitly pass through `YYSTYPE` around.

`%define api.pure full` is preferred to `%pure-parser` (deprecated in
2008, around Bison 2.3b). Should be safe to use.

### DIFF
--- a/lib/lex.h
+++ b/lib/lex.h
@@ -1,13 +1,14 @@
 #ifndef _RE2C_LIB_LEX_
 #define _RE2C_LIB_LEX_
 
+#include "lib/parse.h"
 #include "src/msg/location.h"
 #include "src/regexp/re.h"
 #include "src/parse/ast.h"
 
 namespace re2c {
 
-int lex(const uint8_t*& pattern, Ast& ast);
+int lex(YYSTYPE* yylval, const uint8_t*& pattern, Ast& ast);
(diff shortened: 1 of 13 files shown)


lib/lex.h
```diff
@@ -1,13 +1,14 @@
...
 
+#include "lib/parse.h"
 #include "src/msg/location.h"
 #include "src/regexp/re.h"
 #include "src/parse/ast.h"
 
 namespace re2c {
 
-int lex(const uint8_t*& pattern, Ast& ast);
+int lex(YYSTYPE * yylval, const uint8_t*& pattern, Ast& ast);
```
skvadrik, line 11: Please remove space between type and `*`, as in `YYSTYPE*`. Here and everywhere in the patch.

skvadrik (changes requested):

lib/lex.h
trofi: Sounds good. Dropped leading whitespace. PTAL.

skvadrik (approved):

Issue #327, opened 2020-10-20
Add rst2c tool instead of gen_help.sh.


skvadrik:
Currently the way we generate help message is from the .rst sources (shared between the manpage, help message and online documentation) to an intermediate manpage, and from the intermediate manpage to a C++ literal string: doc/help.rst.in -> (configure) help.rst -> (rst2man) help.rst.1 -> (gen_help.sh) help.cc.

It is possible to simplify the process, bypass the intermediate manpage and generate C++ source code from the .rst sources directly, if we write a tool rst2c that will convert help.rst to help.cc. Then we could get rid of the man invocation and make it work on Windows without Cygwin.

sergeyklay:
I'd like to declare some points that we may want to test during some test suite:

- The manpage was created
- The output of `--help` is not empty
- The line width of `--help`  output is 80
- The `--help`  output starts from USAGE word w/o extra lines before
- The `--help`  output "footer" does not contain extra lines
- The content of  the `--help`  output is the same as the relevant manpage part

sergeyklay:
Just to share one of the possible solutions for testing the command output: for blackbox testing (run a command and verify the output/result) I used https://github.com/chriscool/sharness

skvadrik:
This looks neat!

Comparing command output against the reference output will not work if the .rst sources have changed (because the output will change accordingly). This comparison will only succeed when the help is regenerated from the same sources.

Ideally we should also run spell-checker on all .rst files (with a custom dictionary), such as `aspell -c`. I find errors way too often when I run it manually after updating the docs. But that is a different issue and quite a bit of work on top of writing the script.

idigdoug:
I think this can be closed now.

skvadrik:
> I think this can be closed now, unless we want to wait for the autotools build to match.

We do: the two build systems should be on par, as they are both used. Now that you've done the difficult part and wrote the scripts, it's should be easy to fix autotools and get rid of the old gen_help.sh (I can take a look after merging you PRs https://github.com/skvadrik/re2c/pull/421 and https://github.com/skvadrik/re2c/pull/422).

idigdoug:
Sounds good. I can look into it as well. Agree that it shouldn't be hard.

skvadrik:
@idigdoug fixed this in https://github.com/skvadrik/re2c/pull/419 and https://github.com/skvadrik/re2c/pull/426. Closing.

PR #419, opened 2022-10-23
CMake - Windows-compatible doc bootstrapping

idigdoug:
Pretty much every Python install includes Docutils, and Docutils is 95% of what is needed to bootstrap docs on Windows. Fill in the rest.

I think this addresses issue #327

- `docutils_txt.py` is a Docutils text-output Writer class with support for writing a C file.
- `rst2man.py` is a wrapper for Docutils converting to MAN file. (Copied from Docutils repo.)
- `rst2txt.py` is a wrapper for Docutils converting to TXT file. (Same as rst2man.py but it uses docutils_txt as the Writer.)
- `split_man.py` is a Python implementation of `split_man.sh`.
- Change CMake build to use the new scripts.


idigdoug:
More precisely, it fixes the CMake side of things. I don't have the autoconf stuff running in my environment right now, so I didn't mess with that.

idigdoug:
The help.cc output is a little bit different.
- I didn't indent quite as much as the man page did.
- I put in a few additional empty lines between some things.
- This doesn't try to wrap in the middle of words.

idigdoug:
I reverted the updated bootstraps before committing... Not sure if you wanted them included. Let me know.

build/split_man.py
```diff
@@ -0,0 +1,45 @@
...
+if len(sys.argv) != 4:
+    print('usage:', sys.argv[0], '<input> <output> <lang>')
+    exit(1)
+
+input = sys.argv[1]
+output = sys.argv[2]
+lang = sys.argv[3].lower()
+
+if lang == 'c':
+    lang = b'c'
```
skvadrik, line 14: I wonder if you could just convert `lang` to a byte string upon initialization by appending `.encode('ASCII')` (on line 11) and delete lines 14, 20 and 26.

build/split_man.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
+
```
skvadrik, line 2: Can you add a comment saying `split_man.py ... A script that generates language-specific manpages by substituting a selected set of words in the input manpage` or something like that? Just to be as neat as rst2man.py and rst2text.py.

build/rst2man.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
+
+# Author:
+# Contact: grubert@users.sf.net
+# Copyright: This module has been placed in the public domain.
```
skvadrik, line 5: Generally we don't put license information in every file, but if you want to keep author information, can you unify this header with rst2text.py?

cmake/Re2cGenDocs.cmake
```diff
@@ -5,12 +5,13 @@ function(re2c_gen_manpage source target bootstrap lang)
         add_custom_command(
             OUTPUT "${target}"
             COMMAND "${CMAKE_COMMAND}" -E make_directory ${targetdir}
-            COMMAND "${re2c_splitman}" "${source}" "${source_l}" "${lang}"
-            COMMAND "${RST2MAN}" --tab-width=4 "${source_l}" "${target}"
+            COMMAND "${PYTHON}" "${re2c_splitman}" "${source}" "${source_l}" "${lang}"
+            COMMAND "${PYTHON}" "${re2c_rst2man}" --tab-width=4 "${source_l}" "${target}"
```
skvadrik, line 9: Is `--tab-width=4` used in rst2man.py? I don't see the arguments either used or forwarded.

cmake/Re2cGenDocs.cmake
```diff
@@ -29,12 +30,11 @@ function(re2c_gen_help source target bootstrap)
         add_custom_command(
             OUTPUT "${target}"
             COMMAND "${CMAKE_COMMAND}" -E make_directory ${targetdir}
-            COMMAND "${RST2MAN}" "${source}" "${target}.1"
-            COMMAND "${re2c_genhelp}" "${target}.1" "${target}"
+            COMMAND "${PYTHON}" "${re2c_rst2txt}" --variable-name=help "${source}" "${target}"
```
skvadrik, line 33: Is `--variable-name=help` used in rst2txt.py? I don't see the arguments either used or forwarded.

CMakeLists.txt
```diff
@@ -23,27 +23,6 @@ if(RE2C_REBUILD_PARSERS AND NOT RE2C_FOR_BUILD)
 endif()
 
 option(RE2C_REBUILD_DOCS "Regenerate manpage" OFF)
-if(RE2C_REBUILD_DOCS)
-    find_program(RST2MAN NAMES rst2man rst2man.py)
-    if(NOT RST2MAN)
-        message(FATAL_ERROR "rst2man (or rst2man.py) program is required for RE2C_REBUILD_DOCS")
-    endif()
```
skvadrik, line 30: It should be possible to check for a python docutils package with something like this:
```
execute_process(
    COMMAND Python3::Interpreter -c "import docutils"
    RESULT_VARIABLE EXIT_CODE
    OUTPUT_QUIET
)
if (NOT ${EXIT_CODE} EQUAL 0)
    message(FATAL_ERROR "python package docutils (needed for docs) not found")
endif()
```
(Taken from https://stackoverflow.com/questions/62553007/cmake-check-python3-and-python-packages-presence.)

build/rst2man.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
```
skvadrik, line 1: This needs to be `#!/usr/bin/env @PYTHON@` (and patched after configure step).

See https://github.com/skvadrik/re2c/blob/master/CMakeLists.txt#L99.

build/rst2txt.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
```
skvadrik, line 1: This needs to be `#!/usr/bin/env @PYTHON@` (and patched after configure step).

build/split_man.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
```
skvadrik, line 1: This needs to be `#!/usr/bin/env @PYTHON@` (and patched after configure step).

skvadrik (changes requested):
Thank you for the patch!

A few small things:

- Please upload the updated bootstrap files in a sepate commit on top of this one as part of the same PR, so that I can see the changes (I'll run your PR locally and compare it to the uploaded artifacts to see how stable they are).

- Please add to commit headline that this is only for cmake (something like "CMake: Windows-compatible docs regeneration").

- GithubActions for macOS builds currently fail as https://github.com/skvadrik/re2c/actions/runs/3305730856/jobs/5456773661#step:15:21. Python package docutils _is_ installed (see https://github.com/skvadrik/re2c/blob/master/.github/workflows/ci-cmake.yml, step `Install Python Dependencies`). I'm not sure if it's related to the shebang I asked you to fix in one of the comments.

skvadrik:
@idigdoug Also please enable docs regeneration on CI (https://github.com/skvadrik/re2c/blob/master/.github/workflows/ci-cmake.yml).

build/rst2man.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
+
+# Author:
+# Contact: grubert@users.sf.net
+# Copyright: This module has been placed in the public domain.
+
+"""
+man.py
+======
```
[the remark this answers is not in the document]
idigdoug: I noticed this too. This file is copied from [rst2man.py](https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/tools/rst2man.py). It seemed reasonable to just leave the file as-is rather than make private edits to it.

build/split_man.py
idigdoug: Sure, that would work too.

build/split_man.py
idigdoug: Sounds good.

build/rst2man.py
idigdoug: The rst2man.py is a direct copy from Docutils repo. It seems reasonable to leave it as-is.

I'll remove it from rst2txt.py, as the file is really not worth worrying about.

cmake/Re2cGenDocs.cmake
idigdoug: Command-line parsing and handling is done by the Docutils `publish_cmdline` function. The `-tab-width` option gets forwarded to the rst parser.

cmake/Re2cGenDocs.cmake
idigdoug: Command-line parsing and handling is done by the Docutils `publish_cmdline` function. The `-variable-name` option gets forwarded to the txt writer. See the `settings_spec` variable.

CMakeLists.txt
idigdoug: Done.

build/rst2man.py
idigdoug: This seems messy and unnecessary. It adds an unnecessary step to the build and it puts build artifacts into a non-build folder. You can get the same result by just putting `${PYTHON}` before each call to the script.

idigdoug:
> * GithubActions for macOS builds currently fail as https://github.com/skvadrik/re2c/actions/runs/3305730856/jobs/5456773661#step:15:21. Python package docutils _is_ installed (see https://github.com/skvadrik/re2c/blob/master/.github/workflows/ci-cmake.yml, step `Install Python Dependencies`). I'm not sure if it's related to the shebang I asked you to fix in one of the comments.

The issue is that the Python modules are being installed here:

`/Users/runner/hostedtoolcache/Python/3.10.8/x64/lib/python3.10/site-packages`

But the Python installation found by `find_package` is here:

`/usr/local/Frameworks/Python.framework/Versions/3.10/bin/python3.10`

idigdoug:
> @idigdoug Also please enable docs regeneration on CI (https://github.com/skvadrik/re2c/blob/master/.github/workflows/ci-cmake.yml).

I'm not sure I understand this. It's already building the `-full` preset, and that has `RE2C_REBUILD_DOCS=true`. Am I missing something?

idigdoug:
Sorry, the review re-request was a click.

build/rst2man.py
idigdoug: Also, it doesn't work on Windows, so in all of the build scripts you have to put `${PYTHON}` in front of each use of the script anyway.

skvadrik:
Thanks for addressing the comments! There are a few more things:

- Please squash https://github.com/skvadrik/re2c/pull/419/commits/8653f08f58d7f02b4cf1f833ba813e5cb9fdfa7d into https://github.com/skvadrik/re2c/pull/419/commits/e933043650192eae865f22c330504eff405c0933 before merging

- Also please enable docs regeneration on Windows on CI:

> I'm not sure I understand this. It's already building the -full preset, and that has RE2C_REBUILD_DOCS=true. Am I missing something?

Docs regeneration is currently disabled on Windows builds on CI (see e.g. this build: https://github.com/skvadrik/re2c/actions/runs/3314965984/jobs/5476614508#step:14:17). This is disabled in cmake presets on [line 262](https://github.com/skvadrik/re2c/blob/master/CMakePresets.json#L262), [line 282](https://github.com/skvadrik/re2c/blob/master/CMakePresets.json#L282) and [line 302](https://github.com/skvadrik/re2c/blob/master/CMakePresets.json#L302). After you remove it, docs will be regenerated on CI builds. Also there is [some special-casing in [ci-cmake.yml](https://github.com/skvadrik/re2c/blob/master/.github/workflows/ci-cmake.yml#L116)** for Windows that should probably be removed.

skvadrik:
As for python root dir, I think issue is related: https://github.com/actions/setup-python/issues/162#issuecomment-1238566878 (it hasn't yet been fixed on macOS, which falls in line with the only failing builds being the macOS ones). Let's keep the `-DPython3_ROOT_DIR` workaround for now.

skvadrik:
Tiny thing: on the first commit [re2c - Windows-compatible doc bootstrapping](https://github.com/skvadrik/re2c/pull/419/commits/362ed310c5d66d565827f11c05fe0887a8ce1251), did you mean "CMake: Windows-compatible doc bootstrapping" ? I think it's the last one --- checks are passing.

skvadrik:
Merged. Thanks @idigdoug !

PR #421, opened 2022-10-29
CMake - don't require Python just to build

idigdoug:
The BUILD.md document implies that re2c should be able to build without Python, but the CMake build has a hard requirement for Python. Fix the CMake build to match what BUILD.md says.

- Move the initialization of the PYTHON variable up higher in the script so it is available sooner.
- Only set the PYTHON variable if we find a suitable version of Python.
- If we can't find Python, it's only an error if we are rebuilding docs.
- If we can't find Python, don't add the test or check targets.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,18 +34,21 @@ option(RE2C_REGEN_BENCHMARKS "Regenerate C code for benchmarks" OFF)
 
 # checks for programs
 find_package(Python3 COMPONENTS Interpreter)
-# starting from cmake 3.19 find_package can do version check, but we are on 3.12
-if(Python3_VERSION VERSION_LESS 3.7)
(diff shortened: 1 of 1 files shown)


skvadrik (changes requested):
I think it would be better to always require python and fix BUILD.md (omit "optionally" from "and optionally Python"). There are subtle dependencies on Python in different parts of the project (e.g. there is a benchmarks subdirectory that has some build rules which I'm sure depend on python and don't check for it). But most of all, I don't want to make test targets optional: it would be confusing for someone without python to run `make tests` and find out that no such target exists. Below I suggested a hack for `make check`, but we cannot add such hacks to all targets that are now becoming conditional.

idigdoug:
Makes sense to me. I'll update the BUILD.md pr accordingly.

Issue #423, opened 2022-10-31
Bad codepoint range


ccleve:
I'm trying to compile the exact example here:

https://re2c.org/manual/manual_rust.html#encoding-support

I've copied unicode_categories.re into the test directory and attempted to compile the code using re2rust. I get this error:

`tests/pipeline/tokenizers/re2c/unicode_categories.re:2:68: error: bad code point range: '0xF8 - 0x2C1'`

What am I doing wrong?

Also: I see that unicode_categories.re is three years old at this point. Should it be regenerated with a more recent version of unicode?

skvadrik:
> What am I doing wrong?

Can you provide you command line? Did you forget `--utf8` argument?

> Also: I see that unicode_categories.re is three years old at this point. Should it be regenerated with a more recent version of unicode?

Yes, it should. We even had a project for rewriting the generator (https://github.com/skvadrik/re2c/issues/235#issuecomment-605680011) but that somehow got stuck.

ccleve:
Yes, I did forget the --ut8 argument. Thank you, works now.

I'll take a look at the regeneration code. Maybe I can help.

Issue #235, opened 2018-12-22
Unicode character classes


terpstra:
Firstly, thanks a lot for this tool. It saved me a lot of time! I am using re2c to create a parser for an as-yet unpublished build tool. The input files are utf-8 encoded. Everything works fine for the ascii character set.

However, I'd like to expand my identifier space to include/allow unicode letters in addition to [a-zA-Z]. Currently the only way to do this that I can see is to write a parser for UnicodeData.txt that grabs all of the letter category code points and dumps them into a giant character class. That's fine, but now I have a generator for a generator for C++. It seems like this sort of Unicode character class functionality would be more naturally supported directly in re2c itself.

I was somewhat surprised this was not already supported, so I went looking for these classes in re2c and could not find them. Apologies if this is already supported and my grep-powers were insufficient.

Thanks!

skvadrik:
Hi @terpstra , your grep was correct: re2c doesn't support syntactic aliases for Unicode character classes yet. There is no technical reason it can't do that, but you are the first to ask.

As a temporary quick workaround, I can generate and distribute together with re2c source code an "official" file with re2c definitions of Unicode categories: [unicode_categories.re.txt](https://github.com/skvadrik/re2c/files/2705159/unicode_categories.re.txt). This is to be included verbatim into your .re files; the name `L` can be used in subsequent re2c blocks to denote Unicode letters. The definitions are generated from the same scripts that generate re2c tests, so the definitions are coherent with what re2c is able to handle at the moment. The generator doesn't use UnicodeData.txt directly (though it should), it uses haskell [Data.CharSet](http://hackage.haskell.org/package/charset-0.3.7.1/docs/Data-CharSet.html) library.

terpstra:
Thanks a lot for this! Does re2c support some form of 'include'? Dumping tables this large into a source file whose main focus is parsing distracts the reader.

Ultimately, I think users will want all the classes and subclasses in Unicode. For example, also the Lu class for upper-case letters / etc. Do you think this is a good candidate for future inclusion?

skvadrik:
> Does re2c support some form of 'include'?

No, but it would be useful. Initial implementation may only allow to include files from current directory (the one re2c is run from), otherwise we'd also need to support include paths.

> Ultimately, I think users will want all the classes and subclasses in Unicode.

Agreed.

> Do you think this is a good candidate for future inclusion?

Yes. Don't close the issue. :)

terpstra:
[...]

PR #425, opened 2022-11-02
Add script to regen unicode file

ccleve:
Addresses #235, #423

This seems to work. The format of new file is a little different: the char classes are sorted, and single-char ranges are replaced by just a single char. For example,

[\u1234-\u1234] -> [\u1234]


include/generate_unicode_files.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/python
```
skvadrik, line 1: `#!/usr/bin/env python`

skvadrik (changes requested):
Thanks for taking care of this! A few thigs:

- Please rebase to get rid of the merge commit (should be trivial)

- We need to make sure the script is not installed (I think it is as part of the include subdirectory).

- Loking at the changes in the generated files, they seem to be more substantial than just sorting and single-character classes.
  + The `L_` category (whatever it was) has disappeared; this breaks backwards compatibility.
  + I looked at the beginning of `L` category and I see that range `\u0561-\u0587` is replaced by `\u0560-\u0588`. There may be many more such changes. We need to understand how many and if they are correct --- I suspect they may be already fixed in the haskell [Data.Charset](https://hackage.haskell.org/package/charset-0.3.9/docs/Data-CharSet.html) library that I used to generate old unicode_categories.re Therefore let me first regenerate and commit the files with the [haskell script](https://github.com/skvadrik/re2c/blob/master/test/encodings/unicode_groups.hs) (I will also sort and fix single-character classes), and then let's move to your script. I'll try to do this by the end of today.

- Ideally, it would also be good to generate test files with the same python library as the categories and get rid of the haskell script (not that it doesn't work, but it adds an extra dependency and is generally more difficult for people to run than python). But I can take care of it later.

skvadrik:
I have regenerated unicode_categories.re and tests with the old haskell script: https://github.com/skvadrik/re2c/commit/e3ec259703ab47008b1e4996317ad23b37131921, and I can already see that the nontrivial changes to the character ranges are the same as in the python script. What remains is to sort and fix single-character classes. I'll try to do that ASAP but I'm traveling in the next few days so it may have to wait a bit.

ccleve:
I'm really struggling to figure out how to rebase and get rid of the extra commit. Apparently something I did closed this PR. Git is such a walking disaster...

skvadrik:
> I'm really struggling to figure out how to rebase and get rid of the extra commit. Apparently something I did closed this PR. Git is such a walking disaster...

It can be very confusing. I github auto-closed because you pushed a commit saying "[Merge pull request](https://github.com/skvadrik/re2c/pull/425/commits/fc35c2a797f7fb30de5ca758a1b9581fb1203968) https://github.com/ccleve/re2c/pull/1 [from skvadrik/master](https://github.com/skvadrik/re2c/pull/425/commits/fc35c2a797f7fb30de5ca758a1b9581fb1203968)" (it's a github feature, not a git one). I wonder if it can be configured in settings (to disallow github to be "smart" and close PRs / bugs based on keywords).

So what you need to do now to get nice linear history without merge commits is:
- `git rebase -i HEAD~3`
- in the editor, you will see three latest commits: 1) https://github.com/skvadrik/re2c/pull/425/commits/0a8aa3b084e288c27b129175a40ae67956aa5ca7, 2) [Merge pull request](https://github.com/skvadrik/re2c/pull/425/commits/fc35c2a797f7fb30de5ca758a1b9581fb1203968) https://github.com/ccleve/re2c/pull/1 [from skvadrik/master](https://github.com/skvadrik/re2c/pull/425/commits/fc35c2a797f7fb30de5ca758a1b9581fb1203968) and 3) [Merge branch 'master' of](https://github.com/skvadrik/re2c/pull/425/commits/43fd0be02dcec13a84f065a72ca2bf9e92036bc5) https://github.com/ccleve/re2c.
- You need to squash commits 2 and 3 into 1.
- To do that, replace "pick" word in front of commits 2 and 3 with "fixup" (it is the same as "squash", but it does not try to merge commit messages and uses the first one.
- Save and exit the editor (git should say "successfully rebased...").
- Then `git push -f` (force-push, as you are rewriting history, which is not allowed by default as it is a destructive operation).

And after I merge my remaining work (sort + fix dupes), you will need to rebase your work on top of mine:
- `git pull --rebase skvadrik master`

ccleve:
Thanks, but I'm not seeing the same commits after the git rebase command, and I can't take the time to figure this out.

I recommend just closing or deleting this PR, and then copying and pasting my script into your repo. Life's too short to deal with git nonsense.

On another matter: it seems that it's not easy to get Python to spit out other unicode character properties, like Word Break or Script. I haven't found a good module that can do them, and I don't want to parse unicodedata.txt myself. I ended up writing code in Java to generate these files because  Java and ICU4J support is really, really good. I'm happy to contribute Java code to generate the files, although it will be a lot harder for users to use Java than Python.

skvadrik:
> I recommend just closing or deleting this PR, and then copying and pasting my script into your repo. Life's too short to deal with git nonsense.

Ok, I'll see what I can do. Thanks for the script!

> On another matter: it seems that it's not easy to get Python to spit out other unicode character properties, like Word Break or Script. I haven't found a good module that can do them, and I don't want to parse unicodedata.txt myself. I ended up writing code in Java to generate these files because Java and ICU4J support is really, really good. I'm happy to contribute Java code to generate the files, although it will be a lot harder for users to use Java than Python.

Right, a Java script has the same problem as a Haskell script: it may be nontrivial to run (depending on the developer's environment.

PR #426, opened 2022-11-06
Autotools - use rst2txt for docs

idigdoug:
- Prequisite check for Python3 docutils module instead of rst2man.
- Use `split_man.py` instead of `split_man.sh`.
- Use `rst2txt.py` instead of `rst2man` + `gen_help.sh`.
- Explicitly use `$(PYTHON) <scriptname>` when calling python scripts.


skvadrik:
Thanks!

PR #439, opened 2023-04-20
Add support for http://bazel.build.

helly25:
Bazel can be used as a tool in foreign bazel projects. There is allso
some (yet incomplete) support for developing re2c using bazel.

There are two modes of bazel integration:

1) Using `BUILD.bazel` builds re2c using bazel's actual build support.
This way supports toolchains and with that any compiler or cross
compilation automatically. However, this version assumes some posix
compliant functions and headers.

2) Using the alternative `BUILD.cmake.bazel` file will build re2c using
its cmake build support (`https://github.com/bazelbuild/rules_foreign_cc`).
This option is slower and intended for cases with an existing toolchain.

Limitations:
- Not all documentaton is being generated: Only the one that ends up in
  the binary.
- Only .c based tests are executed.
- No library tests ar implemented yet.


skvadrik:
@helly25 this should be closed now, right?

PR #442, opened 2023-04-25
Provide a min viable http://bazel.build integration.

helly25:
This allows other projects that use bazel to utilize re2c as a tool.

Limitations:

. Input files (docs, lexer, parser) are not being re-generated.
. No tests are executed.
. Only the main binary is being build as nothing else is needed.

A user will load the package using bazel's 'http_archive' in their
WORKSPACE file.

In their BUILD files they can then use:

```
load("@re2c//bazel:re2c.bzl", "re2c_gen")

re2c_gen(
    name = "file_re2c",
    srcs = ["file.re2c"],
    outs = [
        "file.cc",
        "file.h",
    ]
)

cc_library(
    name = "file_cc",
    srcs = ["file.cc"],
    hdrs = ["file.h"],
)
```

The approach here is to automate as much as possible. In that way, all
source and header files will be found automatically (using glob). Also,
the provided bazel integration is kept to an absolute minimum.

An alternative is to use a `rules_foreign_cc` based approach. But that
would require much more bazel work and knowledge. Plus the result would
end up being a non-hermetic build, thereby bypassing bazel's advantages.
Further, using a rules_foreign_cc approach is much slower (for me ~x10).


BUILD.bazel
```diff
@@ -0,0 +1,68 @@
...
+        cat $(location //:CMakeLists.txt) |\\
+            sed -rne 's/project\\(re2c VERSION ([^ ]+).*/#define PACKAGE_VERSION "\\1" RE2C_VERSION_SUFFIX/p' >> "$@"
+        cat $(location //bazel:re2c_config.h) >> "$@"
+    """,
+)
+
+cc_library(
+    name = "re2c_config_cc",
+    hdrs = ["config.h"],
+    defines = ["""RE2C_STDLIB_DIR='"$(BINDIR)/re2c.runfiles/re2c/include"'"""],
```
skvadrik, line 20: Are the single quotes needed? How does expansion work in this case?

bazel/re2c.bzl
```diff
@@ -0,0 +1,81 @@
...
+        default = Label("//:re2c"),
+        allow_single_file = True,
+        executable = True,
+        cfg = "exec",
+    ),
+}
+
+re2c_gen = rule(
+    attrs = _re2c_attr,
+    doc = """Use re2c tool to create an output file (e.g. '.cc') from a '.re2' input.""",
```
skvadrik, line 78: The usual file extension is .re, not .re2.

bazel/re2c.bzl
```diff
@@ -0,0 +1,81 @@
+# re2c generator rules - assuming that re2c is available as '@re2c//:re2c'.
+
+visibility("public")
+
+def _re2c_impl(ctx):
+    srcs = ctx.files.srcs
+    if len(srcs) != 1:
+        fail("Attribute `srcs` must contain exactly 1 '.re2c' file.")
```
skvadrik, line 8: The usual file extension is .re, not .re2c.

bazel/re2c.bzl
```diff
@@ -0,0 +1,81 @@
...
+    outs = ctx.outputs.outs
+    if not outs or [o for o in outs if o.path.split(".")[-1] not in ["c", "cc", "h"]]:
+        fail("Attribute `outs` must contain at least 1 '.cc' or '.h' file.")
```
skvadrik, line 11: Should we also add .hpp and .cpp, perhaps even .cxx?

Also, isn't there a standard function for taking file extension?

bazel/re2c.bzl
```diff
@@ -0,0 +1,81 @@
...
+    err_file = ctx.actions.declare_file(ctx.attr.name + ".err")
+    ctx.actions.run_shell(
+        outputs = outs + [err_file],
+        tools = [ctx.executable._re2c_tool],
+        inputs = ctx.files.srcs + ctx.files.data,
+        command = """
+            line_args=""
+            [ -z "${re2c_parse_first_line}" ] && \\
+                line_args="$(sed -rne "1s,// re2c .INPUT -o .OUTPUT (.+),\\1,p" "{src}")"
```
skvadrik, line 20: I believe this was needed for tests. Is it still needed?

If we want to provide a mechanism to specify re2c command-line options from the first line of a file, I think it should be done in a generic way by adding another re2c option, not by build system hacks. Most of the options already have a corresponding configuration anyway that can be specified in the file.

bazel/re2c.bzl
```diff
@@ -0,0 +1,81 @@
...
+            "{re2c}" "{src}" -o "{out}" ${{line_args}} {re2c_args} {includes} > "{err}" 2>&1
+            cat "{err}" >> "{out}"
```
skvadrik, line 22: Why do you append re2c error messages / warnings to the output file? We do this for tests to make diffing simpler, but I wouldn't expect it in a normal re2c invocation. I'd rather just let them be without redirecting to `{err}` file.

By the way, where does the `{err}` file go, is it just left in the build directory?

skvadrik (changes requested):
Thanks! This is much simpler and more maintainable.

I have a few comments (see below), but my main question is: why do you need to build re2c in such a complex way: split the sources into a few libraries and then build the final binary from them? Now that you only use pre-generated bootstrap files, can't you just grab all the sources and build the executable in one go?

BUILD.bazel
helly25: The single quotes enclose the string that will end up as the definition. So here we get a definition with double quotes which then become a string. Given that this holds a directory there could be spaces and so having the double quotes already here is the best option.

bazel/re2c.bzl
helly25: Done.

bazel/re2c.bzl
helly25: Done.

bazel/re2c.bzl
helly25: Good point. That would be much better anyway - I might send a change for that.

bazel/re2c.bzl
helly25: The error file stays in bazel's gendir output directory. It serves no further use at this point. In theory we could do the same with a temp file, but it is better to let bazel decide where the file goes. And once it is generated, it might we well be kept in place.

In any event it is likely better to do this already in the tool if needed. That will also make it easy to issue "#error" lines.

bazel/re2c.bzl
```diff
@@ -0,0 +1,72 @@
...
+_C_SRC_EXT = [".c", ".cc", ".cpp", ".cxx", ".c++", ".C", ".cu", ".cl"]
+_C_HDR_EXT = [".h", ".hh", ".hpp", ".ipp", ".hxx", ".h++", ".inc", ".inl", ".tlh", ".tli", ".H", ".tcc"]
+
+def _re2c_impl(ctx):
+    srcs = ctx.files.srcs
+    if len(srcs) != 1:
+        fail("Attribute `srcs` must contain exactly 1 '.re' or '.re2c' file.")
+    outs = ctx.outputs.outs
+    if not outs or [o for o in outs if paths.split_extension(o.path)[1] not in _C_SRC_EXT + _C_HDR_EXT]:
+        fail("Attribute `outs` must contain at least 1 '.cc' or '.h' file.")
```
skvadrik, line 19: Nit: the error message is now outdated as it suggests only `'.cc' or '.h'` are accepted. Maybe say something more generic, like `at least one C/C++ source or header file`?

BUILD.bazel
```diff
@@ -0,0 +1,68 @@
...
+        "src/options/symtab.cc",       # Depends on bootstrap sources
+        "src/parse/input.cc",          # Depends on bootstrap sources
+    ]),
+    deps = [
+        "re2c_config_cc",              # Needed for 're2c_config_cc' data files
+    ],
+)
+
+cc_library(
+    name = "re2c_bootstrap_cc",
```
skvadrik, line 46: Still unresolved a question from previous patchset: why do you need to split the sources into a few libraries and then build the final binary from them? Now that you only use pre-generated bootstrap files, can't you just grab all the sources and build the executable in one go?

skvadrik (changes requested):
Thanks! A few remaining questions, mainly about the bootstrap library.

Also please rebase (we dont use merge commits on master) and squash bazel commits into one.

CI failures are related to my yesterday's commit, not to your PR. I'm ivestigating why it failed on Windows, but unfortunately GithubActions CI fails to upload the failed test artifacts (on Windows only) so I wasn't able to fix it quickly.

bazel/re2c.bzl
helly25: Done.

BUILD.bazel
```diff
@@ -0,0 +1,46 @@
...
+genrule(
+    name = "re2c_config",
+    srcs = [
+        "CMakeLists.txt",
+        "//bazel:re2c_config.h",
+    ],
+    outs = ["config.h"],
+    cmd = """
+        cat $(location //:CMakeLists.txt) |\\
+            sed -rne 's/project\\(re2c VERSION ([^ ]+).*/#define PACKAGE_VERSION "\\1" RE2C_VERSION_SUFFIX/p' >> "$@"
```
skvadrik, line 12: Thanks for simplifying. Is RE2C_VERSION_SUFFIX still needed after simplification? Here and on line 38. My understanding was that you needed it to differentiate between `re2c_bootstrap_cc` and the removed library.

skvadrik (changes requested):

BUILD.bazel
helly25: The other aspect is of course that this way I can add 'arbitrary' version suffixes to indicate that this is not a fully correct built version.

skvadrik: > The overall better solution is to have a function that returns the version and have that function be implemented in either main.cc or some new verson.cc/.h file(s). In that case I can remove this here and figure out later what to do with the version. Thoughts?

A function in a separate file sounds good, provided that it will use PACKAGE_VERSION define. There's already a similar one: https://github.com/skvadrik/re2c/blob/master/src/msg/msg.cc#L99-L106, so you can add one with a suitable interface in a separate file and call it from the existing one to avoid duplicating the logic. It will require changes in source code and other build systems to add the new file, so it's better to have as a separate commit/PR.

Then for this PR you can remove `RE2C_VERSION_SUFFIX` logic. Also please rebase and squash bazel-related commits into one, then it'll be good to go.

.gitignore
```diff
@@ -11,6 +11,8 @@
 /m4
 /test_*
 /.build*
+/.vscode/*
```
sergeyklay, line 14: Screenshot 2023-04-26 at 4 06 01 PM

helly25: Done.

skvadrik:
@helly25 Can you rebase and squash all commits except https://github.com/skvadrik/re2c/pull/442/commits/99404d027bb10be710fe4bf6f26d64dced63016f into one?

skvadrik:
Sorry for nitpcking, there is still one merge commit https://github.com/skvadrik/re2c/pull/442/commits/ccf3fe60f8e6500cc01f34516490de59b99e0037, can you get rid of that? If that's hard to do for some reason I'm ok with having one merge commit, but the general rule is to always rebase and keep a simple linear history on master.

helly25:
My bad, I picked the wrong base obviously. Maybe I need to find a UI based
tool that can do these things reliably.

skvadrik (approved):

skvadrik (approved):

skvadrik:
Thanks for addressing all the comments! Good to go (ignoring the three failures on Windows which are my fault).

PR #468, opened 2024-01-31
.github/workflows/ci-cmake.yml: avoid `pwd` and use `${{ github.works…

trofi:
…pace }}`

It loks like at least in `windows` `actions/upload-artifact@v4` CI step fails to collect files at `${{ steps.build-info.outputs.BUILD_DIR }}/test_[0-9]*`

My theory is that `glob` is not able to match paths like

    `D:/a/re2c/re2c/install-2`

against theit unix-like form of

    `/d/a/re2c/re2c/install-2`

I suspect that `pwd` returns us the second form while `${{ github.workspace }}` might keep an original form.

### DIFF
--- a/.github/workflows/ci-cmake.yml
+++ b/.github/workflows/ci-cmake.yml
@@ -135,9 +135,9 @@ jobs:
         id: build-info
         run: |
           if echo "${{ matrix.name }}" | grep -q "ootree"; then
-            BUILD_DIR="$(pwd)/.build/${{ matrix.name }}-full"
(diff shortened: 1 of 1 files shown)


skvadrik:
https://github.com/skvadrik/re2c/actions/runs/7732301081/job/21081937426?pr=468#step:19:26 your theory. :)

skvadrik (approved):

trofi:
Added and removed bad commit. Should be mergeable now.

Issue #450, opened 2023-06-21
Add syntax files to simplify porting re2c to new languages.


skvadrik:
Syntax files should be config files that describe a language backend via a set of configurations. When generating code, re2c would map various codegen concepts to the descriptions provided by the syntax file. This way a new language can be added easily by supplying a syntax file (by the user or by re2c developers --- existing backends should be described via syntax files, distributed with the re2c source code and as part of a re2c installation).

The man difficulty is to decide on a minimal set of configurations that are orthogonal and capable of describing different languages, so that we don't have to add new ad-hoc configurations for each new language. Once this is decided, codegen subsystem should be modified to support syntax files, and exising backends should be rewritten using syntax files (before adding new ones).

Related bugs/commits:
- JS: https://github.com/skvadrik/re2c/issues/420
- OCaml: https://github.com/skvadrik/re2c/issues/449
- Dlang: https://github.com/skvadrik/re2c/commit/742ff817ac9688655c4eb006d266e30fbbafeac9
- HLS: https://github.com/skvadrik/re2c/issues/294

pmetzger:
FWIW, this would be an amazingly cool feature.

pmetzger:
Just wanted to say, again, this would be an amazingly cool feature.

skvadrik:
I started some experimental work on this. I'm constrained on time at the moment, so it's not moving fast, but it's my next most important feature for re2c.

skvadrik:
An update. I've been doing some experimental work on `syntax-file` branch, and I'm now at the point when all three existing language backends (C/C++, Go, Rust) can be expressed via config files: https://github.com/skvadrik/re2c/tree/syntax-files/include/syntax. There `language` option has been removed from re2c, meaning that the codebase is now clear from any language-specific checks: https://github.com/skvadrik/re2c/commit/80bb7db5dca4c7497fd07078b3daf4879ccf7619 (any language-specific behaviour is based on syntax configs). For the three existing backends, config files are built into re2c so that there's not need to pass any syntax configs (the `--lang` option works as before, as well as re2go and re2rust binaries), so this is all backwards compatible.

skvadrik:
Next thing for me is to write a syntax config for D (very close to C/C++) and resurrect test cases from https://github.com/skvadrik/re2c/pull/431.

Feedback on the current DSL I used in syntax file is welcome. Although it's not documented yet, but it shouldn't be very hard to understand. There are different groups of configurations:

- list configurations
- single-word configuration
[...]

PR #469, opened 2024-02-14
add support for V lang

felipensp:
This PR aims to add support for V lang on re2c.

### Example

```V
fn lex(str string) {
	mut cursor := 0
	/*!re2c
		re2c:define:YYCTYPE = u8(0);
		re2c:define:YYPEEK = "str[cursor]";
		re2c:define:YYSKIP = "cursor += 1";
		re2c:yyfill:enable = 0;

		number = [1-9][0-9]*;

		number { return }
		*      { panic("error!") }
	*/
}

fn main() {
	lex("1234\x00")
}
```

### Running
`re2v  ../v_example_01.v  | v run -`


skvadrik:
Hi @felipensp, thanks a lot for your work on this!

It would be great to add V backend, but it needs to be done in a different way. Recently there have been quite a few requests to add other languages (D, Js, OCaml, Zig... now V). The problem with the current approach is, it does not scale well: the code becomes polluted with special cases for each language. Most importantly, in order to test all re2c features with the new code, it is necessary to implement all [examples](https://github.com/skvadrik/re2c/tree/master/examples) and update docs, including https://re2c.org. Understandably, contributors don't have time for all this work.

So, I'm working on a different approach that will make implementing new backends as easy as writing a single config file that describes element of the syntax of the target language. See https://github.com/skvadrik/re2c/issues/450 for details. The work is being done on `syntax-files` branch. I already made good progress on this, and I even have OCaml support locally (which is much more different from C/C++ than V). So I think it will be an easy job to add V after this work is done.

With the new approach, we can have second-tier language support (just add a syntax file, but don't add any tests/docs). It won't affect re2c code.

Unfortunately, I have not much free time at the moment. I'm working on it, but but it's a matter of months, not weeks/days. I should not rush this work anyway, because the moment it gets released it will become stable user interface for the foreseeable future. So I cannot promise I will add V support soon.

Parts of your work can still be reused (those related to options and build systems).

skvadrik:
Vlang support was added in experimental branch `syntax-files` in https://github.com/skvadrik/re2c/commit/73853c5ea7a98d19c174f508bb226876c61a024f.

skvadrik:
In particular, see examples here: https://github.com/skvadrik/re2c/tree/73853c5ea7a98d19c174f508bb226876c61a024f/examples/v

skvadrik:
Closing as obsolete - Vlang support has been merged into master.

felipensp:
Many thanks @skvadrik

PR #470, opened 2024-02-22
[bazel] Fix re2c_gen rule: remove a reference to an undefined variable

egorich239:
Mainly `[err_file]` is an undefined variable. My buildifier also put some random white spaces, I hope it's okay.

Probably some tl&c would be worth it, converting to bazel module, but that's for another day.


skvadrik:
Thanks for the fix!

I would prefer a more descriptive commit message (something like you wrote in the PR description), but ok to leave as is for such a small change.

helly25 (approved):
Nice.

PR #473, opened 2024-03-04
Correct definition of stub CLOSE macro

StephenWassellSophos:


trofi (approved):
Great catch! Thank you!

PR #477, opened 2024-04-24
Use Python version in find_package()

petk:
This simplifies the find_package() call. The version in find_package() has been initially supported in CMake 2 and 3. Version ranges (versionMin...[<]versionMax), on the other hand, are only supported by CMake 3.19+.

And this also helps a bit when adding re2c as a FetchContent dependency in CMake project where finding Python is not necessary and can be disabled with CMAKE_DISABLE_FIND_PACKAGE_Python3 variable.


skvadrik:
Thanks!

PR #481, opened 2024-06-13
Flag to make python optional as build dependency

d-musique:
Hello. The re2c package requires python for docs and testing.
If the test targets can be disabled, the requirement of python can be dropped as well.
This introduces the flag `RE2C_BUILD_TESTS` for this purpose.

The deeper purpose is to facilitate integration of re2c into vcpkg,
which trivializes access to re2c on a Windows development environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,13 @@ include(Re2cCompilerFlags)
 
 ac_subst(PACKAGE_VERSION "${PROJECT_VERSION}")
 
+# check whether re2c is the root project
+if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+    set(RE2C_IS_ROOT_PROJECT TRUE)
+else()
+    set(RE2C_IS_ROOT_PROJECT FALSE)
+endif()
+
 option(RE2C_REBUILD_LEXERS "Regenerate lexers" OFF)
 if(RE2C_REBUILD_LEXERS AND NOT RE2C_FOR_BUILD)
     message(FATAL_ERROR "option RE2C_FOR_BUILD is required for RE2C_REBUILD_LEXERS")
@@ -32,8 +39,13 @@ option(RE2C_BUILD_RE2RUST "Build re2rust executable (an alias for `re2c --lang r
 option(RE2C_BUILD_BENCHMARKS "Build benchmarks" OFF)
 option(RE2C_REGEN_BENCHMARKS "Regenerate C code for benchmarks" OFF)
 
+# test targets are enabled by default only if re2c is the root project
+option(RE2C_BUILD_TESTS "Build tests" "${RE2C_IS_ROOT_PROJECT}")
+
 # checks for programs
-find_package(Python3 3.7 REQUIRED COMPONENTS Interpreter)
+if(RE2C_REBUILD_DOCS OR RE2C_BUILD_TESTS OR RE2C_BUILD_BENCHMARKS)
+    find_package(Python3 3.7 REQUIRED COMPONENTS Interpreter)
+endif()
 
 if(RE2C_REBUILD_DOCS)
     execute_process(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -32,8 +32,20 @@ option(RE2C_BUILD_RE2RUST "Build re2rust executable (an alias for `re2c --lang r
...
+else()
+    set(RE2C_IS_ROOT_PROJECT FALSE)
+endif()
+
+# test targets are enabled by default only if re2c is the root project
+option(RE2C_BUILD_TESTS "Build tests" "${RE2C_IS_ROOT_PROJECT}")
+
 # checks for programs
-find_package(Python3 3.7 REQUIRED COMPONENTS Interpreter)
+if(RE2C_REBUILD_DOCS OR RE2C_BUILD_TESTS)
```
skvadrik, line 46: This should also include `RE2C_BUILD_BENCHMARKS` as they depend on  a few python scripts (to run the benchmarks).

CMakeLists.txt
```diff
@@ -32,8 +32,20 @@ option(RE2C_BUILD_RE2RUST "Build re2rust executable (an alias for `re2c --lang r
 option(RE2C_BUILD_BENCHMARKS "Build benchmarks" OFF)
 option(RE2C_REGEN_BENCHMARKS "Regenerate C code for benchmarks" OFF)
 
+# check whether re2c is the root project
+if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+    set(RE2C_IS_ROOT_PROJECT TRUE)
+else()
+    set(RE2C_IS_ROOT_PROJECT FALSE)
+endif()
```
skvadrik, line 40: I wonder if it can be rewritten more elegantly using generator expressions: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#string-comparisons

skvadrik (changes requested):
Thanks for the patch! It looks good overall, please see a few comments below.

CMakeLists.txt
d-musique: Likely not. I'm pretty certain generator expression don't work in conditionals.
They expand in a later step of the build process.
I'm not aware of a direct way to express this more consisely (that does not require a helper macro).

skvadrik: Ok, thanks for confirming. In that case let's just move this block above line 14, so that options are grouped together.

skvadrik (changes requested):

skvadrik (approved):

skvadrik:
Thanks!

skvadrik:
Can you squash the three commits into one before merging?

d-musique:
> Can you squash the three commits into one before merging?

@skvadrik Sure, yes.

If you allow me a question, also relevant to packaging:
the option `RE2C_BUILD_LIBS` installs a lib currently without headers.

Is this something that is supposed to be accessible to developers by public API?

I need to ask, because it conditions whether package `re2c` must be placed or not in the "tool-only" category, which implies some significant changes.

skvadrik:
> If you allow me a question, also relevant to packaging: the option `RE2C_BUILD_LIBS` installs a lib currently without headers.
> 
> Is this something that is supposed to be accessible to developers by public API?
> 
> I need to ask, because it conditions whether package `re2c` must be placed or not in the "tool-only" category, which implies some significant changes.

libre2c is an experimental library; I'm not aware of it being used in the wild. It exists for the sake of testing and benchmarking various internal regexp-matching algorithms against other libraries that have POSIX interface (we cannot use a lexer generator in that case, as it will be comparing ahead-of-time code generation to just-in-time one, which is meaningless).  The library is not stable. One day it may become useful, but for now I wouldn't bother packaging it, unless you have some specific uses in mind.

d-musique:
Thank you!

PR #485, opened 2024-08-20
Fix my name and change myself from email to github. Also add some history.

helly25:
Fix my name and change myself from email to github. Also add some history.


skvadrik:
Oh, good catch, and sorry about the misspelled name.

PR #494, opened 2024-10-24
Fix static analysis warnings

PolarGoose:
I just checked the code with a static analyzer. Found several small things.
I hope that GitHub actions will test the changes :)

### DIFF
--- a/src/skeleton/generate_data.cc
+++ b/src/skeleton/generate_data.cc
@@ -364,7 +364,9 @@ static void gencover(Skeleton& skel, cover_t& cover, size_t i) {
 
     if (node.end()) {
         suffix.init = true;
-    } if (suffix.init) {
+    }
+
+    if (suffix.init) {
         prefix.push_sfx(suffix);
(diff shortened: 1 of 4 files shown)


src/skeleton/generate_data.cc
```diff
@@ -473,6 +473,7 @@ Ret emit_data(Skeleton& skel) {
     FILE* keys = fopen (keys_name.c_str(), "wb");
     if (!keys) {
         RET_FAIL(error("cannot open file: %s", keys_name.c_str()));
+        fclose(input);
```
skvadrik, line 476: `fclose` needs to go before `RET_FAIL`, as the macro returns.

skvadrik (changes requested):
Thanks!

src/skeleton/generate_data.cc
PolarGoose: Haha, sorry for that.
Fixed.

skvadrik (approved):

PR #499, opened 2024-11-07
[Web playground] Add "Share" button. Add icons for the buttons.

PolarGoose:
I have uncommented `set -o xtrace` line.
Do we really want it to be commented out? Then, it is not clear what happens when scripts are run.

### DIFF
--- a/sync_with_branch.sh
+++ b/sync_with_branch.sh
@@ -3,12 +3,27 @@
 source ./utils.sh
 
 if [[ $# -eq 0 ]]; then
-  echo "Usage: $0 <git-branch-name>"
+  echo "Usage: $0 <git-branch-name> [--no-svg] [--no-playground]"
   exit 1
 fi
 
 branch="$1"
 remote=`git config branch.master.remote`
+
+gen_svg=1
+gen_playground=1
+while [[ $# -gt 1 ]]; do
+    if [[ "$2" == "--no-svg" ]]; then
+        gen_svg=0
+        shift
+    elif [[ "$2" == "--no-playground" ]]; then
+        gen_playground=0
+        shift
+    else
+        echo "Error: unknown argument: $2"
+        exit 1
+    fi
+done
 
 if ! git ls-remote --heads "$remote" "$branch" | grep --quiet "$branch"; then
     echo "branch $branch not found in remote $remote"
@@ -69,11 +84,15 @@ gen_bar_charts_for_benchmark() {
     )
 }
 
-gen_bar_charts_for_benchmark dfa_aot ""
-gen_bar_charts_for_benchmark dfa_jit "--relative-to ''"
-gen_bar_charts_for_benchmark nfa     "--relative-to LG"
+if [[ "$gen_svg" -ne 0 ]]; then
+    gen_bar_charts_for_benchmark dfa_aot ""
+    gen_bar_charts_for_benchmark dfa_jit "--relative-to ''"
+    gen_bar_charts_for_benchmark nfa     "--relative-to LG"
+fi
 
-echo "Build playground"
-playground/build.sh "$branch"
+if [[ "$gen_playground" -ne 0 ]]; then
+    echo "Build playground"
+    playground/build.sh "$branch"
+fi
 
 echo "The script has finished successfully"


playground/index.html
```diff
@@ -38,13 +36,13 @@
...
       display: grid;
-      grid-template-columns: 100px 100px 200px;
+      grid-template-columns: 100px 100px 200px 100px;
       gap: 0.5em;
       padding: 0.5em 0 0.5em 0;
     }
 
     #compileButton {
-      background-color: limegreen;
+      background-color: palegreen;
```
skvadrik, line 45: Regarding the color scheme, I actually would prefer to use re2c.org color scheme. )I know I changed the compile button to lime green, but I planned to amend the colors before pushing). So the editor background should be `aliceblue`, the output `linen` and the compile button shades of `lightsteelblue` (a few shades lighter when not hovered, and exactly `lightsteelblue` when hovered). See the screenshot.
Screenshot(8)

What do you think?

playground/index.html
```diff
@@ -143,25 +142,35 @@
   </div>
 
   <div class="compile-and-open-lexer-graph-layout">
-    <button id="compileButton" onclick="compileCode()">Compile</button>
-    <button onclick="clearLog()">Clear log</button>
-    <button id="openLexerGraphButton" onclick="openLexerGraph()">Open in GraphvizOnline</button>
+    <button id="compileButton" onclick="compileCode()">
+      <img src="icons/GenerateFile.svg" width="16">
```
skvadrik, line 146: TBH I think it's better without the icons. They are rendered differently in different browsers, and in mine they look too small and not aligned with the text. If you want a sign there, maybe use a Unicode symbol? But plain text is also fine.

playground/index.html
```diff
@@ -211,10 +220,23 @@
...
+    const compressedDataBase64 = LZString.compressToEncodedURIComponent(JSON.stringify(
+      {
+        "sourceCode": editor.getValue(),
+        "options": re2cFlagsDropBox.value,
+        "language": languageSelect.options[languageSelect.selectedIndex].text,
+      }
+    ));
+    const shareableURL = `${window.location.origin}${window.location.pathname}?compressed=${compressedDataBase64}`;
+    navigator.clipboard.writeText(shareableURL)
+    window.alert("The share link is copied to clipboard");
```
skvadrik, line 234: This didn't work for me on localhost, presumably because of a non-secure context (the error was `navigator.clipboard is undefined`). I was able to fix it by replacing the highlighted code with this:
```js
    if (navigator.clipboard && window.isSecureContext) {
        navigator.clipboard.writeText(shareableURL)
    } else {
        // Use the 'out of viewport hidden text area' trick
        const textArea = document.createElement("textarea")
        textArea.value = shareableURL

        // Move textarea out of the viewport so it's not visible
        textArea.style.position = "absolute"
        textArea.style.left = "-999999px"

        document.body.prepend(textArea)
        textArea.select()

        try {
            document.execCommand('copy')
        } catch (error) {
            console.error(error)
            alert("Cannot copy shared link!")
        } finally {
            textArea.remove()
        }
    }
    window.alert("Share link is copied to clipboard.")
```

playground/index.html
```diff
@@ -2,10 +2,8 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <link rel="shortcut icon" href="../_static/favicon.ico"/>
-
-  <!-- We don't have a favicon. The browser tries to request a favicon when the page is loaded. Therefore, we need this line to avoid an error being shown in the HTTP server log -->
-  <link rel="icon" href="data:," />
```
skvadrik, line 8: Ugh, I can't read a few lines below the code I'm editing. :)

playground/index.html
```diff
@@ -104,7 +103,7 @@
           <option value="javascript">js</option>
           <option value="ocaml">ocaml</option>
           <option value="python">python</option>
-          <option value="v">v</option>
+          <option value="text">v</option> <!-- There is no syntax highlighting for V lang. Therefore, we use "text" -->
```
skvadrik, line 106: Should we use Go for V? The languages are very similar.

skvadrik (changes requested):
Thanks @PolarGoose, that was fast!

skvadrik:
> I have uncommented set -o xtrace line.
> Do we really want it to be commented out? Then, it is not clear what happens when scripts are run.

I'd prefer it commented out. The general rule for build scripts is to be silent or output only a few "this step is complete" messages on successful execution, and be verbose only in case of errors (and then keep verbosity scoped to the error). This way developers don't have to needlessly look at pages of text, trying to understand why it was printed and if it contains some errors/warnings that require attention. We keep to this rule of thumb in other re2c build /test scripts, build system, etc.

In particular, I run this script quite often when working on the docs, and I'd like to see as small output as possible when I focus on the content changes (not script changes). If you wish, you can add a `--verbose` option (or if you really prefer verbose by default, add a `--quiet` option).

FYI I [added a few options](https://github.com/skvadrik/re2c/commit/dbfb5825ab0ae0c3aea7f76c2ea5aa76fedf54d9) to the sync script.

playground/index.html
PolarGoose: Ok, I have changed the colors so that they match your proposal.

playground/index.html
PolarGoose: Ok, let's just keep plain text.
I have removed the icons.

playground/index.html
PolarGoose: Oh god :)
I have fixed it in a better way. Please check that it works on your machine.

playground/index.html
PolarGoose: What is "V lang" anyway? :)
It seems to me that Go highlighting doesn't work well for this strange unicorn of a language.
I would leave it like that. But we can change if you want.

PolarGoose:
> I'd prefer it commented out. 

Ok, np. If it is more convenient, then it is fine.
I have commented it out like it was before.

playground/index.html
skvadrik: https://vlang.io 

Ok, let's leave it as is.

playground/index.html
skvadrik: Yes, it does. Thanks.

PR #502, opened 2024-11-19
Playground small improvements

PolarGoose:
* I see the Iosevka font was commented out. Do we want to use the default font? Then, we can completely remove it from the page.
* Also I added `.idea/` into .gitignore. It is just for convenience. WebStorm (or any JetBrains IDE) can be used to edit the playground website.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build/
 obj/
 re2c.org/
+/temp_build_playground/
--- a/playground/index.html
+++ b/playground/index.html
@@ -2,14 +2,12 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
+
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
   <link rel="icon" href="../_static/favicon.ico"/>
 
   <title>re2c Online. Version:</title>
-
-  <!-- There is no specific reason why this font is chosen. The main rationale is that it is condensed and allows longer lines to be visible on the screen -->
-  <link href="https://iosevka-webfonts.github.io/iosevka/Iosevka.css" rel="stylesheet"/>
 
   <style>
     body {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 build/
 obj/
 re2c.org/
+.idea/
+/temp_build_playground/
```
skvadrik, line 5: Ignore files specific to the user belong in a user-wide ignore file (https://git-scm.com/docs/gitignoreб see `$GIT_DIR/info/exclude`).

playground/index.html
```diff
@@ -196,14 +197,14 @@
   const editor = ace.edit("editor")
   editor.setTheme("ace/theme/github_light_default");
   editor.setOptions({
-    // fontFamily: "Iosevka Web",
+    fontFamily: "Iosevka Web",
```
skvadrik, line 200: I wrote an explanation in commit that commented these out: https://github.com/skvadrik/re2c/commit/885b665cfc109a7d0d4b7dcb1ad712298fcf1232:
```
Although Iosevka is a nice font, it may be a bit hard to read as it's very
narrow. It is better to use default system or browser monospace font, which
should be familiar to the user (and this is what we do for the examples on
re2c.org anyway).
```

I didn't delete them in case we change our mind later. If you don't like to leave them as comments, feel free to remove.

PolarGoose: > Although Iosevka is a nice font, it may be a bit hard to read as it's very
narrow.

You haven't seen [Quinze font](https://www.programmingfonts.org/#quinze) :)
But fair enough. I have removed Iosevka. The default font is good enough.

playground/index.html
```diff
@@ -66,7 +66,6 @@
     }
 
     #editor, #output, #log-output {
-      border: 1px solid lightgray;
```
skvadrik, line 69: Why is this removed?

PR #508, opened 2024-11-22
Playground: add "example" URL parameter to specify the example

PolarGoose:
Changes:
* add "example" URL parameter to specify the example
* update re2c.js with the latest master commit
* fix typos in the build.sh script

How to use the URL parameter:
* `re2c.org/playground/?example=c/fill/01_fill.re`
* `re2c.org/playground/?example=c/01_basic.re`

Could you please test that it works?


playground/index.html
```diff
@@ -93,19 +93,15 @@
...
 
     input, select {
-      font-family: mono, 'Iosevka Web';
       margin: 0 0.5em 0 0.5em;
     }
 
-    textarea {
-      font-family: mono, 'Iosevka Web';
-    }
-
```
skvadrik, line 103: This is actually needed: it sets font `mono` (default monospace) before Iosevka (which is a fallback option). Can you remove just Iosevka part?

skvadrik (changes requested):
Can you change commit headline from `Playground:` to `Playground: add URL query parameter to select example` (or something like this, to be more descriptive).

skvadrik:
> Could you please test that it works?

Works like charm! Thanks a lot.

playground/index.html
PolarGoose: np, returned `mono`

PolarGoose:
Fixed the review remarks.

Glad that it works :)

skvadrik (approved):

Issue #379, opened 2021-11-02
Pre-built release binaries


nightlark:
Would there be interest in having a CI workflow that builds (statically linked) Windows/macOS/Linux binaries for re2c/re2go and uploads them when a GitHub release is made?

The re2c binary seems like the main way of using re2c, so having a central place to grab the latest release binaries for common platforms would be nice instead of waiting for some of the update cycles for package managers (older Ubuntu releases being stuck on old versions of re2c, chocolatey package for Window being a few years outdated, etc).

skvadrik:
Agreed, that would be useful. I think @sergeyklay wanted to do that as well.

sergeyklay:
I apologize for completely forgetting this.  My bad. I'll try to sort out asap.

skvadrik:
No worries @sergeyklay, If you haven't started yet, maybe @nightlark was planning to work on this? Just trying to avoid duplicate effort.

nightlark:
Yea I was thinking of starting on this, but I didn't find existing discussion about it on GitHub.

@sergeyklay if you already started something I can see about finishing it if you'd like. Otherwise, I can probably get a first pass at a workflow ready around this upcoming weekend.

sergeyklay:
Unfortunately, I will not be able to find a free time in the next 4-5 days. And no, unfortunately I didn't even start.

That's what I think:

- Every CI run should produce binaries
- We should store these binaries in GitHub Artifacts
- On release event we should get latest binaries from GitHub Artifacts and attach theme to a release

@skvadrik thoughts?

skvadrik:
I'm a bit worried if storing binaries on every CI run will waste too much space. We have `retention-days: 30` but still those binaries won't be used. Maybe upload them only on release?

Aside from that, sounds good.

nightlark:
An option to store the artifacts from a CI run could be added as a separate feature, and/or a nightly/weekly run of the release binary workflow could minimize the space used (I doubt re2c would run into artifact storage limits either way).

Would it be desirable to follow the bootstrapping process to regenerate the .re files before building the final release binaries?

skvadrik:
> Would it be desirable to follow the bootstrapping process to regenerate the .re files before building the final release binaries?

[...]

PR #509, opened 2024-11-23
Increase the Cmake minimum version to allow using CMAKE_MSVC_RUNTIME_LIBRARY flag

PolarGoose:
In Cmake 3.15 introduces CMAKE_MSVC_RUNTIME_LIBRARY
https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html
This flag is very useful to build a static binary on Windows by using `CMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.15)
 project(re2c VERSION 4.0 HOMEPAGE_URL "https://re2c.org/")
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.15)
```
skvadrik, line 1: How do you use CMAKE_MSVC_RUNTIME_LIBRARY? I'm not against increasing, but it's strange to increase it for something that is not used anywhere in the build files.

PolarGoose: It will be needed for this issue: [Pre-built release binaries? #379](https://github.com/skvadrik/re2c/issues/379)
Currently it is used in my repo that produces a Windows binary: [re2c-for-Windows/build.ps1](https://github.com/PolarGoose/re2c-for-Windows/blob/main/build.ps1#L51)

PR #510, opened 2024-11-28
Playground: add example URL parameter when an example is selected

PolarGoose:


skvadrik:
Oh wow that was fast! Thanks a lot @PolarGoose !

skvadrik:
BTW I changed it slightly: https://github.com/skvadrik/re2c/commit/e9a706633ce31b2ea7bc9218759f046b2c476238, because I realized that I have to click "back" more than once (e.g. three times after I follow an URL that has a query string with example). It seems to work well now.

PR #511, opened 2024-11-28
Playground: don't change URL when shared link is used

PolarGoose:
Before this fix, if you follow a shared link, it works properly but the URL gets changed.


skvadrik:
Thanks!

PR #520, opened 2024-12-30
Implement PHP support

Jeckerson:


skvadrik (changes requested):
Thanks for the port!

A minor thing, but can you put all PHP entries sorted alphabetically, between OCaml and  Python?

Also, please rebase on top of master.

This is a partial review. I'll do the full review as soon as I can, but I broke my arm a few days ago, and I'm slower than usual.

Jeckerson:
@skvadrik I adjusted alphabetically the implementation. There are still work in examples/ and I'll need your help in some already made examples, especially with `null` (`[\x00]`) value. As in PHP it must be generated like that: `case "\0":` and not `case 0x00:`.

skvadrik:
> I adjusted alphabetically the implementation.

> There are still work in examples/ and I'll need your help in some already made examples, especially with null ([\x00]) value. As in PHP it must be generated like that: case "\0": and not case 0x00:.

I think you should use `re2c:yych:literals = hex;`, similar to Python that also has no character type (single-quoted literals are strings, as in PHP).

include/syntax/php
```diff
@@ -0,0 +1,381 @@
...
+
+
+// language-specific options ---------------------------------------------------
+
+semicolons = 0;
+backtick_quoted_strings = 1;
+single_quoted_strings = 1;
+indentation_sensitive = 0;
+wrap_blocks_in_braces = 1;
+
```
skvadrik, line 17: I've added a new configuration `special_escapes` that needs to go here, see other syntax files.

examples/php/01_basic.re
```diff
@@ -0,0 +1,16 @@
+<?php
+// re2php $INPUT -o $OUTPUT
```
skvadrik, line 2: I think it should always be the first line of the file (as expected by run_tests.py).

include/syntax/php
```diff
@@ -0,0 +1,381 @@
+// supported feature lists -----------------------------------------------------
+
+supported_apis = ["simple", "generic", "record"];
+supported_api_styles = ["free-form"];
+supported_code_models = ["loop-switch"];
+supported_targets = ["code", "dot"];
+supported_features = ["nested-ifs", "bitmaps", "tags", "captures", "captvars"];
```
skvadrik, line 7: For all supported and unsupported features there must be a test. Tags, captures and captvars are covered by examples, and for the rest see tests in `test/codegen/<language>`.

skvadrik (changes requested):
A few more comments.

skvadrik:
So I downloaded the patch and tried to build some of the problematic examples, but it seems that the .php files are not generated from the current syntax file (it was copied from javascript without modification). Please let me know when you update it (or if you have any questions).

examples/php/01_basic.re
Jeckerson: Impossible, as `<?php` must be on 1st line.

skvadrik: I see. Then I suggest adding some sed magic in `examples/php/__run_all.sh` that will cut off the comments in the temporary file (this one and the one autogenerated by re2php).

PR #528, opened 2025-01-31
migrate to Bazel 8 using bzlmod

mobileink:
This turns out to be trivial, no changes needed other than adding a MODULE.bazel file with a bazel_dep for skylib, and deletion of the WORKSPACE file.

If you want to support previous versions of Bazel you'd have to tell the user to use `--enable_bzlmod` or some such.

BTW thanks for making this available.  I've been using version 2.0.3 for a few years, it's pretty amazing. I've been building it with Bazel and rules_foreign_cc, which usually took over a minute to compile. WIth the latest version and Bazel it take seconds. Yay!


WORKSPACE
helly25, line 1: You need to leave the `WORKSPACE` file alone but add a new empty `WORKSPACE.bzlmod` file. The new empty file can have a license or other comments but must otherwise be empty. The effect is that is someone uses workspaces (old solution) they will work with the old file(s). If someone enables bazel-mod, then they will use the new files. Thus everyone is happy. Deleting `WORKSPACE` means no one can continue using the repo if they are on the old workspace solution.

MODULE.bazel.lock
helly25, line 1: You do should exclude this file from being merged. In that case it will be created on the receiver/user side as needed. Best way is to simply ignore it in .gitignore.

helly25 (changes requested):

MODULE.bazel.lock
mobileink: Depends. If you want replicability, you include the lockfile.

WORKSPACE
mobileink: Yes. Something like that.

MODULE.bazel.lock
mobileink: "Just like go.sum, you should commit your MODULE.bazel.lock file to source control to ensure everyone on your project gets the same version of each dependency. You shouldn't need to edit MODULE.bazel.lock manually."

helly25: Right. And for larger projects with lots of dependencies where specific APIs and stability of said APIs are important that is absolutely the best guidance. However, for highly portable projects like re2c we in fact prefer to get input from as many configurations as possible, plus we anyway control the exact versions we need, meaning we already over specify the dependencies. You can compare us in some ways to Google's Abseil project that also aims at portability and does not merge the lock file (though it is not a tool project but rather a library). That project also does not merge the lock file. Further, if used with Bazel, there is a likelihood that re2c is used as a dependency in a larger bazel project in which case some interactions and some version divergence have to work. Bottom line, for the moment I prefer us to not merge the lock file. Depending on feedback we can go either way as needed and drop precise version specs or rather merge the lock file.

mobileink: Done

helly25 (approved):

skvadrik:
Thanks for your work @mobileink !

mobileink:
Thanks for re2c!  BTW, it would be nice to have re2c in the [BCR](https://registry.bazel.build/). I've never submitted but would be happy to help if you want to do that.

Issue #531, opened 2025-02-08
-g is unfriendly to shared libraries and PIEs


crrodriguez:
https://gcc.gnu.org/onlinedocs/gcc/Labels-as-Values.html

-g should generate code using the "alternate way" documented in the GCC manual to make the generated code friendly with shared libraries and also safer since it is gonna become read only data. It will probably run faster too on most modern systems.

skvadrik:
Thanks for reporting this! It seems that there is one incompatibility (a quote from https://gcc.gnu.org/onlinedocs/gcc/Labels-as-Values.html):

> This alternative with label differences is not supported for the AVR target, please use the first approach for AVR programs.

We can definitely add another option and configuration though.

crrodriguez:
It could however be guarded with ifndef __AVR__ or something similar..

skvadrik:
Link to the GCC bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119158 for the record.

There's nothing we can do in re2c at the moment, so closing the issue (relative `goto` are implemented now).

PR #533, opened 2025-02-20
Relative computed goto (experimental)

caffe3:
I noticed #531 and I thought I'd give it a go because I've been working on some of my own code generation improvements.  This is purely experimental.

I have only done the lightest amount of testing so far (it compiles and the c++98.re seemed to work) and it has some caveats both in my current implementation and compiler support:

- I have had to use typeof to keep C++ happy without introducing another type configuration variable to re2c
- Clang's heuristics do not work very well with the relative offsets

@crrodriguez would you give this a spin to see if it works in your use-case?

To use this, add to a block, and enable `-g` for re2c:

re2c:cgoto:relative = 1


crrodriguez:
Great I'll give a try..

crrodriguez:
unfortunately gnu AS seems to choke on it.. I tried with the PHP parsers.. chokes on var_unserializer parser..I'll see if it is worth reporting..

caffe3:
Hi @crrodriguez - thanks, what sort of errors do you see (are they parsing errors or memory errors)? I'll have some time tomorrow to test it further.

caffe3:
This seems to have an interesting effect on executable size (gcc -Os):

810592 Feb 21 07:51 uri_rfc3986 (re2c 4.0.2 -g)
101912 Feb 21 07:51 uri_rfc3986_2 (re2c draft-cgoto-relative -g --syntax usc) # replaced yytargetoffset with "static const unsigned short")
146968 Feb 21 07:52 uri_rfc3986_3 (re2c draft-cgoto-relative -g) # yytargetoffset with "static const int")

Maybe this is really worth investigating further?

src/codegen/pass2_generate.cc
```diff
@@ -919,19 +937,35 @@ static void emit_accept(
     // jump table
     if (opts->computed_gotos && nacc >= opts->computed_gotos_threshold && !have_tags) {
         CodeList* block = code_list(alc);
+        uint32_t min_index = UINT32_MAX;
```
skvadrik, line 940: Please don't use a shadowed variable (the first definition on line 789), reuse the outer one.

src/codegen/pass2_generate.cc
```diff
@@ -919,19 +937,35 @@ static void emit_accept(
...
 
         const char** elems = alc.alloct<const char*>(nacc);
+        const char* name = opts->var_computed_gotos_table.c_str();
+
         for (uint32_t i = 0; i < nacc; ++i) {
-            buf.cstr("&&").str(opts->label_prefix).u32(acc[i].state->label->index);
-            elems[i] = buf.flush();
+            buf.cstr("(typeof(*").cstr(name).cstr("))((char *)&&").str(opts->label_prefix).u32(acc[i].state->label->index);
+            if (opts->computed_gotos_relative) {
+                buf.cstr(" - (char *)&&").str(opts->label_prefix).u32(min_index);
```
skvadrik, line 954: Let's drop the space betweeb `char` and `*`.

test/config/config7c.re
```diff
@@ -0,0 +1,15 @@
+// re2c $INPUT -o $OUTPUT -g
```
skvadrik, line 1: Did you forget to commit the .c file for this test?

skvadrik (changes requested):
Thanks for your work @caffe3 ! I have a few comments, see below.

skvadrik:
> This seems to have an interesting effect on executable size (gcc 14.2.0 with -Os):
> 
> 810592 Feb 21 07:51 uri_rfc3986 (re2c 4.0.2 -g)
> 101912 Feb 21 07:51 uri_rfc3986_2 (re2c draft-cgoto-relative -g --syntax usc) # replaced yytargetoffset with "static const unsigned short")
> 146968 Feb 21 07:52 uri_rfc3986_3 (re2c draft-cgoto-relative -g) # yytargetoffset with "static const int")
> 
> Maybe this is really worth investigating further?

It is! Since we know all the offsets, we should be able to find the smallest integer type for the table. The change in binary size seems drastic, I'm surprized to see it.

caffe3:
> It is! Since we know all the offsets, we should be able to find the smallest integer type for the table. The change in binary size seems drastic, I'm surprized to see it.

The same thought came to mind but I think that would only work with C++ if at all, unfortunately.  But for most small/medium-sized parsers uint16_t should be sufficient if the compiler doesn't decide to rearrange the earliest label, otherwise int16_t is another reasonable choice.

skvadrik:
> The same thought came to mind but I think that would only work with C++ if at all, unfortunately. But for most small/medium-sized parsers uint16_t should be sufficient if the compiler doesn't decide to rearrange the earliest label, otherwise int16_t is another reasonable choice.

Ah, right, I forgot we're not on the assembly level and we don't really know the size of the generated code and how far the labels will be. Other options that we can consider:

- an estimate of the function size based on the number of states
- a configuration to set `yytarget` type
- assertions to make sure `yytarget` is not overflown

This is all follow-up work that should not get in the way of the initial implementation, though.

BTW why is it different between C and C++?

caffe3:
> BTW why is it different between C and C++?

I was wondering if it was possible to use C++'s constexpr magic to determine the size of the offset values but it isn't possible after some tests - it doesn't treat && as constantexpr values.

crrodriguez:
Regenerating the zend_language_parser.l from PHP with this options results in a parser that compiles but crashes. .. needs further investigation.

caffe3:
> Regenerating the zend_language_parser.l from PHP with this options results in a parser that compiles but crashes. .. needs further investigation.

Thanks for testing! Could you give me the version of PHP + compiler you used to build it and any other environment information that could potentially affect it?  And some commands that can reproduce the crash - I'm not terribly familiar with php.

src/parse/conf_lexer.re
```diff
@@ -271,6 +271,7 @@ Ret Input::lex_conf(Opt& opts) {
     "flags:"? "debug-output"   | "flags:d"    { RET_CONF_BOOL(debug); }
     "flags:"? "computed-gotos" | "flags:g"    { RET_CONF_FEAT(computed_gotos, "computed-gotos"); }
     ("computed-gotos" | "cgoto") ":threshold" { RET_CONF_NUM_NONNEG(computed_gotos_threshold); }
+    ("computed-gotos" | "cgoto") ":relative"  { RET_CONF_BOOL(computed_gotos_relative); }
```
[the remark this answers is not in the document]
caffe3: Should this become `"flags:" ? "computed-gotos-relative" | "flags:r" { RET_CONF_FEAT(computed_gotos_relative, "computed-gotos-relative"); }`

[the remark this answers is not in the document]
skvadrik: > Should this become `"flags:" ? "computed-gotos-relative" | "flags:r" { RET_CONF_FEAT(computed_gotos_relative, "computed-gotos-relative"); }`

No, no need for the deprecated `flags:...` form for new configurations. Also, `-r` is taken anyway (short for `--reusable`) and I didn't mean to add short option for this one, long option is enough.

src/codegen/pass2_generate.cc
caffe3: Fixed

src/options/opt.cc
```diff
@@ -196,6 +196,7 @@ LOCAL_NODISCARD(Ret fix_mutopt(
     }
     if (!real.computed_gotos) {
         real.computed_gotos_threshold = defaults.computed_gotos_threshold;
+        real.computed_gotos_relative = defaults.computed_gotos_relative;
```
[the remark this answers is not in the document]
caffe3: Fixed.

src/codegen/pass2_generate.cc
```diff
@@ -754,19 +754,26 @@ static CodeList* gen_goswif(
...
     const opt_t* opts = output.block().opts;
     OutAllocator& alc = output.allocator;
     Scratchbuf& buf = output.scratchbuf;
 
     const char** elems = alc.alloct<const char*>(CodeGoCpTable::TABLE_SIZE);
+    const char* name = opts->var_computed_gotos_table.c_str();
+
     for (uint32_t i = 0; i < CodeGoCpTable::TABLE_SIZE; ++i) {
-        elems[i] = buf.cstr("&&").str(opts->label_prefix).u32(go->table[i]->label->index).flush();
+        buf.cstr("(typeof(*").cstr(name).cstr("))((char *)&&").str(opts->label_prefix).u32(go->table[i]->label->index);
```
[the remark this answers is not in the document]
caffe3: I'll also probably need to put this function somewhere so that it is cached if this route is okay.

skvadrik (changes requested):

src/codegen/pass2_generate.cc
```diff
@@ -754,24 +756,101 @@ static CodeList* gen_goswif(
...
+    std::string word;
+    while (iss >> word) {
+        buf.cstr(first ? "" : " ");
+        if (!sc.count(word)) {
+            buf.str(word);
+            first = false;
+        }
+    }
+    return buf.flush();
+}
```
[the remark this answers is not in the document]
caffe3: Hi skvadrik - it's to remove storage-specifier like "static" which code:type_yytarget may contain.  When casting `static` isn't valid i.e.

`int abc = (static const int)123;`

Will error.

An alternative approach could be to encode in the syntax file:
```
code:type_yytarget = "int"
code:type_qualifers_yytarget = "static const"
```

or

```
code:type_yytarget = "static const int"
code:type_cast_yytarget = "int"
```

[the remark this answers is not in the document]
skvadrik: Ah, I see. Still, there's no point in adding any complex logic like this in codegen, as it should be all moved to syntax files. The right solution would be to make `code:type_yytarget` free of modifiers, and move modifiers as parameters of generic array construction. But it is too much work unrelated to this PR. Please just hardcode the cast type for now with a TODO comment, it will be easier to port to syntax files than if you add this clever logic.

include/syntax/c
```diff
@@ -4,7 +4,7 @@ supported_apis = ["simple", "generic", "record"];
 supported_api_styles = ["functions", "free-form"];
 supported_code_models = ["goto-label", "loop-switch", "recursive-functions"];
 supported_targets = ["code", "dot", "skeleton"];
-supported_features = ["nested-ifs", "bitmaps", "computed-gotos", "case-ranges",
+supported_features = ["nested-ifs", "bitmaps", "computed-gotos", "computed-gotos-relative", "case-ranges",
```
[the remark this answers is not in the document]
caffe3: Ah okay - I think I originally misunderstood that the features and re2c arguments were not treated the same.

I don't think I've done it correctly, could you help me with `src/options.opt.cc` ?

src/codegen/pass2_generate.cc
```diff
@@ -754,24 +756,101 @@ static CodeList* gen_goswif(
...
+                    buf.cstr("(").str(cast_type).cstr("))((char*)");
+                }
+                buf.str(converted);
+                if (opts->computed_gotos_relative) {
+                    buf.cstr(" - (char*)").str(base_address);
+                    buf.cstr(")");
+                }
+            }
+            return buf.flush();
+        });
```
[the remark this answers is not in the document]
caffe3: Sure, that makes sense - I made it a simpler loop but still in a separate function. I'll also rework the original code I wrote to match.

test/config/config7c.c
```diff
@@ -0,0 +1,30 @@
+/* Generated by re2c 4.1 on Sat Feb 22 23:10:50 2025 */
```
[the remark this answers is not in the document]
caffe3: Threshold has been added so a table is generated now.

src/codegen/pass2_generate.cc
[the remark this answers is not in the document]
caffe3: Okay, that sounds good - I've removed `gen_yytarget_cast_type` and hardcoded to `int`

src/codegen/pass2_generate.cc
```diff
@@ -754,24 +756,101 @@ static CodeList* gen_goswif(
            : gen_goif(output, dfa, go->goif, from);
 }
 
-static CodeList* gen_gocp_table(Output& output, const CodeGoCpTable* go) {
+// TODO: generalise when ported to other langs; this is very C/C++ specific
+static std::string gen_yytarget_cast_type(Scratchbuf& buf, const char *type)
+{
```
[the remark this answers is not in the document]
caffe3: Removed as per comments below.

src/codegen/pass2_generate.cc
```diff
@@ -754,24 +756,78 @@ static CodeList* gen_goswif(
            : gen_goif(output, dfa, go->goif, from);
 }
 
-static CodeList* gen_gocp_table(Output& output, const CodeGoCpTable* go) {
+template<typename iter_t, typename func_t>
+static Code* gen_gocp_code(Output& output, const char* name, iter_t first, iter_t last,
+                           func_t gen_address, const std::string& base_address, bool local = true) {
```
skvadrik, line 761: I think you can simplify things if you don't move the whole array generation into this helper function, but just the subexpression for a single array element. This will allow you to keep most of the previous code in place and still avoid code duplication (which want to avoid, as table elements become more complex as we add the relative form).

test/config/config7c.c
```diff
@@ -0,0 +1,59 @@
...
+// re2c $INPUT -o $OUTPUT -g
+
+#line 6 "<stdout>"
+{
+	YYCTYPE yych;
+	if (YYLIMIT <= YYCURSOR) YYFILL(1);
+	yych = *YYCURSOR;
+	{
+		static const int yytarget[256] = {
+			(int))((char*)&&yy1 - (char*)&&yy1), (int))((char*)&&yy1 - (char*)&&yy1), (int))((char*)&&yy1 - (char*)&&yy1), (int))((char*)&&yy1 - (char*)&&yy1), (int))((char*)&&yy1 - (char*)&&yy1), (int))((char*)&&yy1 - (char*)&&yy1), (int))((char*)&&yy1 - (char*)&&yy1), (int))((char*)&&yy1 - (char*)&&yy1),
```
skvadrik, line 12: This code ends up being so verbose that we might want to generate a local macro for a single element (define it before the table and undef after it). What do you think?

skvadrik (changes requested):

src/options/opt.h
```diff
@@ -457,6 +457,7 @@ enum class StxVarId : uint32_t {
     MUTOPT(std::string, var_accept, "yyaccept") \
     MUTOPT(std::string, var_bitmaps, "yybm") \
     MUTOPT(std::string, var_computed_gotos_table, "yytarget") \
+    MUTOPT(std::string, var_computed_gotos_offset_table, "yytargetoffset") \
```
skvadrik, line 460: No need for this anymore.

include/syntax/c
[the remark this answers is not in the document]
skvadrik: Features are just some selected options or groups of related options. There's no strict correspondence between them.

I think what you do in opt.cc is fine (set `computed_gotos` as implied by `computed_gotos_relative`). Are you concerned about something in partucular?

src/codegen/pass2_generate.cc
caffe3: I think I'll drop the generalisation patch for now to keep the changes down to a more manageable amount

src/options/opt.h
caffe3: Removed this.

test/config/config7c.c
caffe3: I'll make this in a separate PR

src/codegen/pass2_generate.cc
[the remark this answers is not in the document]
caffe3: Hardcoded to "int" until future fixes.

src/codegen/pass2_generate.cc
```diff
@@ -754,19 +754,31 @@ static CodeList* gen_goswif(
...
+
     for (uint32_t i = 0; i < CodeGoCpTable::TABLE_SIZE; ++i) {
-        elems[i] = buf.cstr("&&").str(opts->label_prefix).u32(go->table[i]->label->index).flush();
+        if (opts->computed_gotos_relative) {
+            buf.cstr("(int)("); // TODO: port to syntax file
+            buf.cstr("(char*)");
+            buf.cstr("&&").str(opts->label_prefix).u32(go->table[i]->label->index);
+            buf.cstr(" - (char*)");
+            buf.cstr("&&").str(opts->label_prefix).u32(min_index);
+            buf.cstr(")");
```
skvadrik, line 771: Formatting issue: can you just glue it together into one statement, breaking the lines at 100 char limit (the way it's formatted everywhere else)? The comment should go on the line before the statement and say something like `` TODO: port this to syntax files and replace hardcoded `int` with `code:type_yytarget` ``.

src/codegen/pass2_generate.cc
```diff
@@ -754,19 +754,31 @@ static CodeList* gen_goswif(
...
+        } else {
+            buf.cstr("&&").str(opts->label_prefix).u32(go->table[i]->label->index);
+        }
+        elems[i] = buf.flush();
     }
 
     const char* name = opts->var_computed_gotos_table.c_str();
     const char* type = opts->gen_code_type_yytarget(buf);
-    // In rec/func mode tables are reused across different fnctions, so they must be global.
+
```
skvadrik, line 780: Newline added by accident?

src/codegen/pass2_generate.cc
```diff
@@ -919,19 +946,44 @@ static void emit_accept(
...
         for (uint32_t i = 0; i < nacc; ++i) {
-            buf.cstr("&&").str(opts->label_prefix).u32(acc[i].state->label->index);
+            if (opts->computed_gotos_relative) {
+                buf.cstr("(int)("); // TODO: port to syntax file
+                buf.cstr("(char*)");
+                buf.cstr("&&").str(opts->label_prefix).u32(acc[i].state->label->index);
+                buf.cstr(" - (char*)");
+                buf.cstr("&&").str(opts->label_prefix).u32(min_index);
+                buf.cstr(")");
```
skvadrik, line 966: Same comment as for lines 766-771.

src/codegen/pass2_generate.cc
```diff
@@ -1611,9 +1671,20 @@ static CodeList* gen_cond_table(Output& output) {
...
-    for (size_t i = 0; i < conds.size(); ++i) {
-        elems[i] = buf.cstr("&&").str(opts->cond_label_prefix).str(conds[i].name).flush();
+
+    for (uint32_t i = 0; i < conds.size(); ++i) {
+        if (opts->computed_gotos_relative) {
+            buf.cstr("(int)("); // TODO: port to syntax file
+            buf.cstr("(char*)");
+            buf.cstr("&&").str(opts->cond_label_prefix).str(conds[i].name);
+            buf.cstr(" - (char*)").str(opts->cond_label_prefix).str(conds.front().name);
+            buf.cstr(")");
```
skvadrik, line 1681: Same comment as for lines 766-771.

src/options/opt.cc
```diff
@@ -270,7 +273,8 @@ LOCAL_NODISCARD(Ret fix_mutopt(
     if (real.api_style == ApiStyle::FREEFORM && !glob.supported_api_styles_contains("free-form")) {
         RET_FAIL(error("free-form API style is not supported for this backend"));
     }
-    if (!is_default.computed_gotos && !glob.supported_features_contains("computed-gotos")) {
+    if ((!is_default.computed_gotos || !is_default.computed_gotos_relative) &&
+        !glob.supported_features_contains("computed-gotos")) {
```
skvadrik, line 277: Oh, good that that you noticed this. Can you make it `if (!(is_default.computed_gotos && is_default.computed_gotos_relative) && ...` which is a bit easier to read?

skvadrik (changes requested):
Looks good, only a few minor issues, plus I think you forgot to include regenerated bootstrap files (syntax files, lexers, docs).

src/codegen/pass2_generate.cc
caffe3: Are the following lines just indented by 4 or aligned?

src/options/opt.cc
caffe3: Yep!

src/codegen/pass2_generate.cc
skvadrik: Use 4-space indent (the way you've done now is good). I'm not sure what you mean by aligned though.

skvadrik (approved):

src/codegen/pass2_generate.cc
caffe3: Great, thanks! I meant if the '.' on each line were aligned.

skvadrik:
> I took a look at this again.. it is a bad interaction between this code generation and always_inline functions.. alays_inline function gets ..well..always inlined and vanishes from the resulting asm..

@crrodriguez Did you try with the final state of this PR that has been merged? I think @caffe3 fixed some error and tested that it works with php.

caffe3:
> I think @caffe3 fixed some error and tested that it works with php.

I haven't found the time to test with PHP yet - PHP's build system looked more complicated than I originally thought rather than just clone & configure & build.

But @crrodriguez - if you haven't already pulled the latest, please do, because there was a bug in the initial implementation which would have caused a segfault.

caffe3:
Hi @crrodriguez - I reproduced your issue.  I strongly suspect GCC has a bug where it is still reordering functions into new sections despite being used within relative gotos.

I fixed it by disabling reordering with `-fno-reorder-functions`:

```m4
diff --git a/configure.ac b/configure.ac
index 01d9ded69b..6675e802b7 100644
--- a/configure.ac
+++ b/configure.ac
@@ -155,13 +155,13 @@ PHP_PROG_PHP([7.4])
 PHP_ARG_ENABLE([re2c-cgoto],
   [whether to enable computed goto extension with re2c],
   [AS_HELP_STRING([--enable-re2c-cgoto],
-    [Enable re2c -g flag to optimize conditional jumps using computed goto
+    [Enable re2c --computed-gotos-relative flag to optimize conditional jumps using computed goto
     extension, if supported by the compiler])],
   [no],
   [no])

 AS_VAR_IF([PHP_RE2C_CGOTO], [no],,
-[AC_CACHE_CHECK([whether re2c -g works], [php_cv_have_re2c_cgoto],
+[AC_CACHE_CHECK([whether re2c --computed-gotos-relative works], [php_cv_have_re2c_cgoto],
   [AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
 int main(void)
 {
@@ -176,7 +176,7 @@ label2:
   [php_cv_have_re2c_cgoto=yes],
   [php_cv_have_re2c_cgoto=no])])
 AS_VAR_IF([php_cv_have_re2c_cgoto], [yes],
-  [AS_VAR_APPEND([RE2C_FLAGS], [" -g"])])
+  [AS_VAR_APPEND([RE2C_FLAGS], [" --computed-gotos-relative"])])
 ])

 dnl Platform-specific compile settings.
@@ -201,6 +201,9 @@ esac
 dnl See https://github.com/php/php-src/issues/14140
 AX_CHECK_COMPILE_FLAG([-ffp-contract=off], [CFLAGS="$CFLAGS -ffp-contract=off"])

+dnl Because it breaks re2c --computed-gotos-relative otherwise
+AX_CHECK_COMPILE_FLAG([-fno-reorder-functions], [CFLAGS="$CFLAGS -fno-reorder-functions"])
+
 dnl Mark symbols hidden by default if the compiler (for example, gcc >= 4)
 dnl supports it. This can help reduce the binary size and startup time.
 AX_CHECK_COMPILE_FLAG([-fvisibility=hidden],
```

I ran `make test` and got:

```txt
PASS Test malformed watchpoint name [sapi/phpdbg/tests/watch_007.phpt]
=====================================================================
TIME END 2025-03-05 08:03:47

=====================================================================
TEST RESULT SUMMARY
---------------------------------------------------------------------
Exts skipped    :    43 (bcmath, bz2, calendar, com_dotnet, curl, dba, dl_test, enchant, exif, ffi, ftp, gd, gettext, gmp, intl, ldap, mbstring, mysqli, mysqlnd, odbc, openssl, pcntl, pdo_dblib, pdo_firebird, pdo_mysql, pdo_odbc, pdo_pgsql, pgsql, readline, shmop, skeleton, snmp, soap, sockets, sodium, sysvmsg, sysvsem, sysvshm, tidy, xsl, zend_test, zip, zlib)
Exts tested     :    26
---------------------------------------------------------------------

Number of tests : 20449             14604
Tests skipped   :  5845 ( 28.6%) --------
Tests warned    :     0 (  0.0%) (  0.0%)
Tests failed    :     0 (  0.0%) (  0.0%)
Expected fail   :     8 (  0.0%) (  0.1%)
Tests passed    : 14596 ( 71.4%) ( 99.9%)
---------------------------------------------------------------------
Time taken      : 338.225 seconds
=====================================================================
```

skvadrik:
> I strongly suspect GCC has a bug where it is still reordering functions into new sections despite being used within relative gotos.

I thought relative gotos are good precisely because they don't care about the absolute location of the function, as they are relative to the start of the function. So why would function order matter here?

If you suspect a compiler bug, can you get a self-contained C/C++ file so we can compile it with/without `-fno-reorder-functions`, compare the generated code and hopefully minimize it and report a bug to GCC?

caffe3:
> I thought relative gotos are good precisely because they don't care about the absolute location of the function, as they are relative to the start of the function. So why would function order matter here?

I think the flag is a misnomer - GCC puts different parts of the code into different .text sections to optimise whether its a hot/cold/unlikely path.  I speculate that this lets the linker potentially reorder the functions later.

The changes in the assembly are very small: https://gist.github.com/caffe3/dccc0f1908dcdc9b2f39c371c102ac46

> If you suspect a compiler bug, can you get a self-contained C/C++ file so we can compile it with/without `-fno-reorder-functions`, compare the generated code and hopefully minimize it and report a bug to GCC?

Yes I think that is possible - the assembly currently stands at around 40k lines... and I think only gets triggered if the code of the function is big enough though.

caffe3:
Unfortunately there's no per-function flag that I can see helps with this instance.  I tried `no_reorder, noclone, noinline, hot` and none of them prevent the incorrect assembly, except:

`static __attribute__((cold)) int php_var_unserialize_internal(UNSERIALIZE_PARAMETER)`

crrodriguez:
> > I took a look at this again.. it is a bad interaction between this code generation and always_inline functions.. alays_inline function gets ..well..always inlined and vanishes from the resulting asm..
> 
> @crrodriguez Did you try with the final state of this PR that has been merged? I think @caffe3 fixed some error and tested that it works with php.

Yes, I tried with the revision that was merged or master branch. not using zend_always_inline worked around the problem for me with gcc15, I also believe this to be a compiler bug.

caffe3:
The GCC documentation states:

> The &&foo expressions for the same label might have different values if the containing function is inlined or cloned. If a program relies on them being always the same, __attribute__((__noinline__,__noclone__)) should be used to prevent inlining and cloning. If &&foo is used in a static variable initializer, inlining and cloning is forbidden. 

What did you do to avoid `zend_always_inline` ? It isn't an attribute on the affected function, `php_var_unserialize_internal`.

skvadrik:
> I think the flag is a misnomer - GCC puts different parts of the code into different .text sections to optimise whether its a hot/cold/unlikely path. I speculate that this lets the linker potentially reorder the functions later (edit: it does, and it partitions functions as well!).

> The GCC documentation states:
> 
> > The &&foo expressions for the same label might have different values if the containing function is inlined or cloned. If a program relies on them being always the same, **attribute**((**noinline**,**noclone**)) should be used to prevent inlining and cloning. If &&foo is used in a static variable initializer, inlining and cloning is forbidden.

We don't rely on `&&foo` expressions being the same _always_, we just rely on them being the same within one instance of the re2c-generated function. Is GCC splitting the function, then inlining / cloning some parts of it, so that effectively `&&foo` expressions become different within the same function? If so, it does look like a bug.

caffe3:
Removing `__attribute__((cold))` from the below makes it compile:

```c
__attribute__((cold)) php_error_docref();
php_var_unserialize() {
  static yytarget_0 = (char *) && yy14 - (char *) && yy1;
  goto *&&yy1 + yytarget_0;
yy1:
yy2:
  goto yy2;
yy14:
  php_error_docref();
}
```

PR #535, opened 2025-02-26
Generalize the computed-gotos code generation for C/C++ backend

caffe3:
Previously the code-generation did similar table generation across the normal states, accept states and condtional variables. This refactor introduces some utility functions that provides the same logic across the different tables.

### DIFF
--- a/src/codegen/output.h
+++ b/src/codegen/output.h
@@ -41,6 +41,8 @@ class Scratchbuf {
     Scratchbuf& u64(uint64_t u) { os << u; return *this; }
     Scratchbuf& str(const std::string& s) { os << s; return *this; }
     Scratchbuf& cstr(const char* s) { os << s; return *this; }
+    Scratchbuf& label(const std::string& s) { os << s; return *this; }
+    Scratchbuf& label(uint32_t u) { os << u; return *this; }
     Scratchbuf& exact_uint(size_t width);
     const char* flush();
 
--- a/src/codegen/pass2_generate.cc
+++ b/src/codegen/pass2_generate.cc
@@ -754,6 +754,34 @@ static CodeList* gen_goswif(
            : gen_goif(output, dfa, go->goif, from);
 }
 
+template<typename label_t>
+static const char* gen_gocp_entry(Scratchbuf& buf, bool relative, const std::string& prefix,
+        const label_t base_label, const label_t label) {
+    if (relative) {
+        // TODO: port to syntax file and replace hardcoded `int` with `code:type_yytarget`
+        buf.cstr("(int)(").cstr("(char*)").cstr("&&").str(prefix)
+            .label(label).cstr(" - (char*)").cstr("&&")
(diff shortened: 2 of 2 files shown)


src/codegen/output.h
```diff
@@ -41,6 +41,8 @@ class Scratchbuf {
     Scratchbuf& u64(uint64_t u) { os << u; return *this; }
     Scratchbuf& str(const std::string& s) { os << s; return *this; }
     Scratchbuf& cstr(const char* s) { os << s; return *this; }
+    Scratchbuf& value(const std::string& s) { os << s; return *this; }
+    Scratchbuf& value(uint32_t u) { os << u; return *this; }
```
skvadrik, line 45: Let's call it `label` rather than generic `value` to avoid confusion what method to use for strings / 32-bit ints. 

The idea of explicit methods for different types in `Scratchbuf` was to avoid having one generic method (similar to `<<` operator) which would suit all types and perform sometimes unexpected conversions. The way it is done now is clumsy, but at least you know what you are trying to output (as you have to type it yourself).

src/codegen/pass2_generate.cc
```diff
@@ -754,6 +754,37 @@ static CodeList* gen_goswif(
            : gen_goif(output, dfa, go->goif, from);
 }
 
+template<typename label_t>
+static const char* gen_gocp_entry(Scratchbuf& buf, bool relative, const std::string& prefix,
+                                  const label_t base_label, const label_t label)
```
skvadrik, line 759: Formatting issue: can you make indent 8 or 4 spaces (not aligned) for arguments that don't fit on the same line. This is inconsistent across the codebase, I'm afraid.

Also, can you put the curly brace on the same line (egyptian style). Here and below.

While at it, can you please add a missing newline before line 1660.

skvadrik (changes requested):
A few small issues; looks good overall. Thank you!

src/codegen/output.h
caffe3: Ah that makes sense!

src/codegen/pass2_generate.cc
caffe3: Sure! I'll tweak my editor and environment so I don't run into these issues as often.

src/codegen/pass2_generate.cc
```diff
@@ -754,6 +754,35 @@ static CodeList* gen_goswif(
...
+    } else {
+        buf.cstr("&&").str(prefix).label(label);
+    }
+    return buf.flush();
+}
+
+template<typename label_t>
+static const char* gen_gocp_goto(Scratchbuf& buf, bool relative, const std::string& prefix,
+        const label_t base_label, const std::string& table,
+        const std::string& entry) {
```
skvadrik, line 774: Minor issue, but it seems to fit on the previous line.

skvadrik (changes requested):

skvadrik (approved):
Thanks!

PR #537, opened 2025-03-01
Add code:yytarget_elem to syntax files.

caffe3:
`code:yytarget_elem` specifies the type used for the element types in the computed goto tables.


include/syntax/c
```diff
@@ -185,6 +185,8 @@ code:type_uint = "unsigned int";
 code:type_yybm = "static const unsigned char";
 code:type_yytarget = "static const " (.cgoto.relative ? "int" : "void*");
 
+code:yytarget_elem = (.cgoto.relative ? "int" : "void*");
+
```
skvadrik, line 189: To make it generic, it should be an expression that describes the whole element of the table, not just a type (and it should be added at the end of file, not in the types subgroup). If relative, it should include cast and subtraction, otherwise just the address of a label. The code template will need some local variables (one for the label for which you can use existing variable `label`, another one for the base label which we can call `base` - this one you'll need to add to `RE2C_STX_LOCAL_VARS`).

skvadrik (changes requested):

include/syntax/c
skvadrik: Let me know if you need any help with the code templates language - it's a custom DSL  described here: http://re2c.org/manual/manual_c.html#syntax-files (in the "Code templates" subsection).

caffe3: Thanks! I think I figured it out.

caffe3:
Should I make another syntax variable for relative gotos as well i.e. `gen_gocp_goto` ?

skvadrik:
> Should I make another syntax variable for relative gotos as well i.e. `gen_gocp_goto` ?

You mean another syntax configuration for the goto statement that uses table elements? Yes please. Something like `code:yytarget_goto = topindent <here goes the definition...> nl;`

include/syntax/c
```diff
@@ -441,3 +441,5 @@ code:yylessthan =
 code:yybm_filter = yych " & ~0xFF";
 
 code:yybm_match = yybm "[" offset "+" yych "] & " mask;
+
+code:yytarget_elem = (.cgoto.relative ? "(int)((char*)&&" : "&&") label (.cgoto.relative ? " - (char*)&&" base ")");
```
skvadrik, line 445: I think it will be a bit cleaner if you totally split the two cases, like this:
```
code:yytarget_elem = (.cgoto.relative
    ? "(int)((char*)&&" label " - (char*)&&" base ")"
    : "&&" label);
```

src/codegen/pass2_generate.cc
```diff
@@ -177,6 +177,27 @@ class GenEnumElem : public RenderCallback {
     FORBID_COPY(GenEnumElem);
 };
 
+class GenYytargetElem : public RenderCallback {
```
skvadrik, line 180: Just `GenTargetElem` to match the rest of the names that don't have `YY`.

src/codegen/pass2_generate.cc
```diff
@@ -177,6 +177,27 @@ class GenEnumElem : public RenderCallback {
...
+
+  public:
+    GenYytargetElem(
+            std::ostream& os, const std::string& base, const std::string& label)
+        : os(os), base(base), label(label) {}
+
+    void render_var(StxVarId var) override {
+        switch (var) {
+        case StxVarId::BASE: os << base; break;
+        case StxVarId::LABEL: os << label; break;
```
skvadrik, line 193: Formatting: in a simple switch where each case fits on one line we do indent cases.

src/codegen/pass2_generate.cc
```diff
@@ -755,17 +776,14 @@ static CodeList* gen_goswif(
...
-            .label(label).cstr(" - (char*)").cstr("&&")
-            .str(prefix).label(base_label).cstr(")");
-    } else {
-        buf.cstr("&&").str(prefix).label(label);
-    }
-    return buf.flush();
+    const opt_t* opts = output.block().opts;
+    Scratchbuf& buf = output.scratchbuf;
+    const std::string base = buf.str(prefix).label(base_label).flush();
+    const std::string lbl = buf.str(prefix).label(label).flush();
```
skvadrik, line 784: Instead of creating these temporary strings with unnecessarily long lifetime, can you save options and label references in `GenTargetElem` and output them directly?

skvadrik (changes requested):
Thanks! Looks good overall, just a few suggestions how to polish it.

caffe3:
> > Should I make another syntax variable for relative gotos as well i.e. `gen_gocp_goto` ?
> 
> You mean another syntax configuration for the goto statement that uses table elements? Yes please. Something like `code:yytarget_goto = topindent <here goes the definition...> nl;`

Do you mean to have something like:

`code:yytarget_goto = topindent "goto " (.cgoto.relative ? "*((char*)&&" base " + " array "[" index "])" : "*" array "[" index "]") nl ";";`

skvadrik:
> Do you mean to have something like [...]

Yes, but please use `yytarget` variable instead of `array` like we do for `yybm` and other primitives.

caffe3:
> > Do you mean to have something like [...]
> 
> Yes, but please use `yytarget` variable instead of `array` like we do for `yybm` and other primitives.

I chose array because the code currently overlaps condition table generation with `yytarget`.  I guess then we should consider make a separate set of syntax for `yyctable` (which would be useful for relative gotos anyway so we don't have to use the same types for both of them).  Do you have any naming preferences for conditions?

skvadrik:
> I chose array because the code currently overlaps condition table generation with `yytarget`. I guess then we should consider make a separate set of syntax for `yyctable` (which would be useful for relative gotos anyway so we don't have to use the same types for both of them). Do you have any naming preferences for conditions?

Ah, I see why you chose `array`. But if we'd have to use something more generic than `code:yytarget_elem`, since it would be strange to use `yytarget` in configuration name, but not in the variable name. I see two options.

First, sticking with `yytarget` in both places. As for `yyctable`, making another set of configurations for it seems like a waste: it's a very niche use case, and only for one language. We can reuse `yytarget` configurations for `yyctable` (and substitute `yyctable` for `yytarget` variable) with a comment explaining this in the codegen code. This way we leave the options open for later if we want to add separate configs for `yyctable`.

Second, we can rename `code:yytarget_goto` and `code:yytarget_elem` to `code:cgoto` and `code:cgoto_target` respectively, move them in the syntax file under `code:goto` definition and use `array` variable name. This is more generic and can be reused for any APIs.

What do you think?

skvadrik:
> I think the first option could be useful - there's a high chance that when conditions are used, there will be very few conditions, one could in theory set the elem type for conditions as `uint8_t`.

Ok, let's go this way. Should we also add `code:type_yyctable` then? Could be done in a follow-up commit / PR.

caffe3:
> > I think the first option could be useful - there's a high chance that when conditions are used, there will be very few conditions, one could in theory set the elem type for conditions as `uint8_t`.
> 
> Ok, let's go this way. Should we also add `code:type_yyctable` then? Could be done in a follow-up commit / PR.

Ended up adding it to this commit, if that's okay.

src/codegen/pass2_generate.cc
```diff
@@ -177,6 +177,59 @@ class GenEnumElem : public RenderCallback {
     FORBID_COPY(GenEnumElem);
 };
 
+class GenTargetElem : public RenderCallback {
```
skvadrik, line 180: Should we name them `GenCGotoTableElem` and `GenCGoto` instead of `GenTargetElem` and `GenTargetGoto`, as they are also used for `yyctable`?

skvadrik (changes requested):
There's a failing test on CI: https://github.com/skvadrik/re2c/actions/runs/13864044228/job/38799019483?pr=537. Does it reproduce locally?

skvadrik:
> Ended up adding it to this commit, if that's okay.

Sure, it's fine. Thanks!

src/codegen/pass2_generate.cc
caffe3: Sure!

caffe3:
> There's a failing test on CI: https://github.com/skvadrik/re2c/actions/runs/13864044228/job/38799019483?pr=537. Does it reproduce locally?

I'll investigate that

skvadrik (approved):

skvadrik:
> > First, sticking with `yytarget` [...]
> > Second, [...]
> > What do you think?
> 
> I think the first option could be useful - there's a high chance that when conditions are used, there will be very few conditions, one could in theory set the elem type for conditions as `uint8_t`.

FYI I reworked it: https://github.com/skvadrik/re2c/commit/52fa88db7b3bd53f999533c80d2d6a941588916c (following some prior work to get rid of qualifiers in the type). See the commit message for details.

PR #540, opened 2025-03-23
swift: Support yyfn:throw

ScrelliCopter:
Adds support for yyfn:throw to the Swift syntax, allowing the user to throw types that conform to the Error protocol within functions generated by the recursive-functions code model. Non-empty settings of the `re2c:yyfn:throw` property are appended to generated function definitions between parameters and the return type, and generated function calls are prefixed with "try ".

The 'test/codegen/swift/recursive_functions/advanced.re' test has been updated to test the new functionality as well as demonstrate practical use of throwing functions in the recursive-functions model for handling runtime errors.


skvadrik:
Merged, thank you!

PR #575, opened 2026-04-24
Use std::max_align_t instead of max_align_t.

ryandesign:
This fixes the following build failure on OS X 10.9 and earlier when using contemporaneous versions of Apple clang:

```
In file included from ../src/options/parse_opts.re:6:
In file included from ./src/dfa/dfa.h:12:
In file included from ./src/dfa/tcmd.h:10:
./src/util/allocator.h:85:71: error: unknown type name 'max_align_t'; did you mean 'std::max_align_t'?
constexpr size_t ALLOCATOR_ALIGNMENT = (sizeof(void*) == 4) ? alignof(max_align_t) : sizeof(void*);
                                                                      ^~~~~~~~~~~
                                                                      std::max_align_t
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/cstddef:55:21: note: 'std::max_align_t' declared here
typedef long double max_align_t;
                    ^
```

### DIFF
--- a/src/util/allocator.h
+++ b/src/util/allocator.h
@@ -3,7 +3,7 @@
 
 #include <stdint.h>
 #include <stdlib.h>
-#include <stddef.h>
+#include <cstddef>
 #include <algorithm>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


skvadrik:
This breaks linux builds, see for example https://github.com/skvadrik/re2c/actions/runs/24876351762/job/72833760518?pr=575#step:9:62:
```
./src/util/allocator.h:85:76: error: ‘max_align_t’ is not a member of ‘std’; did you mean ‘max_align_t’?
```

skvadrik:
The breakage on linux is apparently fixed by including `<cstddef>` instead of `<stddef.h>`. Please include this in your patch so that the presubmit checks pass. I'll create a bug to clean up all C++-vs-C header includes for later.

skvadrik:
> I'll create a bug to clean up all C++-vs-C header includes for later.

https://github.com/skvadrik/re2c/issues/576

ryandesign:
Oh! Sorry for the breakage and thanks for suggesting the fix, which is now in this PR and the checks now pass.

skvadrik:
Thanks!

PR #580, opened 2026-07-07
Draft: zig: add computed-continue optimization for loop-switch

PongPong:
## Summary

  Adds --computed-continue, an optimization for the --loop-switch code model. It enables switch jump-table transitions using value-carrying continue statements, similar to --computed-gotos.

  The initial implementation supports Zig’s labeled switch syntax:
  
```
  yystate: switch (@as(enum {
      yy0,
      yy1,
  }, .yy0)) {
      .yy0 => {
          continue :yystate .yy1;
      },
      .yy1 => {
          // ...
      },
  }
```

  ## Implementation

  - Adds the --computed-continue option and validates its dependency on --loop-switch.
  - Adds the global syntax-file conditional .computed_continue.
  - Extends the generic code:loop and code:continue templates with the state information needed by syntax files.
  - Keeps language-specific enum, labeled-switch, and continuation syntax in the Zig syntax file.
  - Reuses the existing loop-switch transition generation for ordinary and computed continuation.
  - Uses the configurable re2c:variable:yystate value as the switch label.
  - Supports custom DFA label prefixes without introducing fixed state names.
  - Updates documentation and generated bootstrap files.
  
  ## Restrictions

  --computed-continue:

  - Requires the --loop-switch code model.
  - Cannot be combined with start conditions.
  - Cannot be combined with storable state.
  - Requires support from the selected backend; currently Zig provides the necessary syntax.

  These constraints are validated centrally in opt.cc, so invalid combinations produce explicit errors instead of silently
  changing behavior.


skvadrik:
Thanks for the pull request, and sorry for the delayed feedback. Can this be implemented using loop/switch code model instead? You could add a thin syntax file that overrides only the necessary configurations to change `--loop-switch` behaviour. I'd rather not add a whole new code model for just one backend, especially since it seems very similar to the existing ones.

PongPong:
Thanks for your comments! I’m still getting to know this repository, so I’ll need a bit more time to dive into the codebase. I’ll give it a go.

skvadrik:
Is this ready for review?

PongPong:
@skvadrik Yes, it’s ready for review now. I’ve addressed the compatibility cases, updated the documentation, and added coverage for EOF/fill transitions, tags, custom labels, invalid option combinations, and user-defined State collisions. The focused fixtures and Zig runtime tests pass. Thanks!

skvadrik:
> @skvadrik Yes, it’s ready for review now. I’ve addressed the compatibility cases, updated the documentation, and added coverage for EOF/fill transitions, tags, custom labels, invalid option combinations, and user-defined State collisions. The focused fixtures and Zig runtime tests pass. Thanks!

Thanks, it looks much better now! The general approach is good. I think it can simplified even further by replacing syntax-file configuration `code:continue_value` with a syntax-file conditional `.continue_value` and reusing existing configuration `code:continue` (just making its implementation conditional depending on `.continue_value`). Then you can get rid of all extra handling for the new syntax configuration. I'll leave detailed comments on the diff.

doc/manual/basics/options/options.rst_
```diff
@@ -141,6 +141,14 @@
     case label of the destination state and continue execution to the head of
     the loop.
 
+``--loop-switch-label -L``
```
skvadrik: Please don't add the short option `-L`: we've only have them for major features and not for all of them, so it would be illogical to have one for a variation of existing feature.

doc/manual/basics/options/options.rst_
```diff
@@ -141,6 +141,14 @@
...
+    Modify ``--loop-switch`` to encode DFA as a labeled switch expression,
+    where transitions use ``continue :label value`` to move between switch
```
skvadrik: I believe it should be called `--computed-continue` similar to the C/C++ backend `--computed-gotos` option, since it has exactly the same purpose. The new option should depend on the `--loop-switch` option.

Also, please drop the short option: we've only have them for major features (or for historically early features).

The description should be something like "This option only works with ``--loop-switch`` code model. It enables optimization of ``switch`` statements known as "jump table", similar to ``--computed-gotos`` option."

Drop the part about Zig and incompatibility with other features: its dependence on `--loop-switch` and other restrictions should be encoded in opt.cc so that the users get error messages if they try to combine incompatible features.

doc/manual/basics/syntax_files.rst_
```diff
@@ -585,6 +585,10 @@ language configurations, inplace configurations and code templates.
   ``.code_model.loop_switch``
       True if code model based on loop/switch is used (``--loop-switch``).
 
+  ``.loop_switch_label``
+      True if loop/switch model uses labeled-switch transitions
+      (``--loop-switch-label``).
+
```
skvadrik: Please make it a global syntax-file conditional `.computed_continue` (similar to `StxGOpt::CGOTO_RELATIVE`).

include/syntax/zig
```diff
@@ -220,13 +220,16 @@ code:switch_case_range = [val{0}: val] (.many ? "..." [val{-1}: val]);
 
 code:switch_case_default = "else";
 
-code:loop =
-    topindent (.loop_label ? label ": ") "while (true) {" nl
+code:loop = (.loop_switch_label
```
skvadrik: Use conditional `.computed_continue` here.

include/syntax/zig
```diff
@@ -220,13 +220,16 @@ code:switch_case_range = [val{0}: val] (.many ? "..." [val{-1}: val]);
...
+    ? topindent label ": " [stmt: stmt]
+    : topindent (.loop_label ? label ": ") "while (true) {" nl
         indent [stmt: stmt] dedent
-    topindent "}" nl;
+    topindent "}" nl);
 
 code:continue = topindent "continue" (.loop_label ? " :" label) ";" nl;
 
+code:continue_value = topindent "continue :" label " " val ";" nl;
+
```
skvadrik: Don't add a new configuration; instead, use conditional `.computed_continue` in `code:continue` above and add the `val` variable there (it will only be available under `.computed_continue` conditional, otherwise it should have some default value).

src/codegen/pass2_generate.cc
```diff
@@ -488,6 +488,21 @@ static void gen_continue_yyloop(Output& output, CodeList* stmts, const char* nex
...
+    return o.flush();
+}
+
+static void gen_continue_yyswitch(Output& output, CodeList* stmts, const Label& next) {
+    const opt_t* opts = output.block().opts;
+    const char* label = opts->label_loop.empty() ? nullptr : opts->label_loop.c_str();
+    append(stmts, code_continue_value(
+            output.allocator, label, gen_switch_label_state(output, next)));
+}
+
```
skvadrik: Reuse `gen_continue_yyloop` instead, conditioning on the feature in its implementation (the fact that you have to place an identical `if/else` at every call site is a hint that the the condition should be moved inside of the function). Rename it to `gen_continue_yyswitch` (or how you see fit) if needed.

src/codegen/code.h
```diff
@@ -245,6 +245,11 @@ struct CodeCgoto {
     const char* index;
 };
 
+struct CodeContinueValue {
+    const char* label;
+    const char* value;
+};
+
```
skvadrik: Rename to just `CodeContinue` (if it's still needed after other modifications)?

src/codegen/pass1_analyze.cc
```diff
@@ -795,9 +795,9 @@ LOCAL_NODISCARD(Ret codegen_analyze_block(Output& output)) {
         case CodeModel::LOOP_SWITCH:
             // First state label is always used, as there are no jumps in the middle of a case.
             dfa->head->label->used = true;
-            // With loop/switch there are no labels, and each block has its own state switch where
-            // all conditions are joined. Restart state counter from zero so that cases start from
-            // zero. With skeleton conditions are separate.
+            // With switch-based models there are no labels, and each block has its own state
+            // switch where all conditions are joined. Restart state counter from zero so that
+            // cases start from zero. With skeleton conditions are separate.
```
skvadrik: Drop this - the comment is fine as is. "Loop/switch" is the name of the code model that used with the new option, so it's fine.

src/codegen/pass2_generate.cc
```diff
@@ -1099,9 +1126,13 @@ static void emit_rule(Output& output, CodeList* stmts, const Adfa& dfa, size_t r
...
-            // loop/switch mode: set `yystate` to the start DFA state of the next condition and
-            // continue to the head of the loop.
-            gen_continue_yyloop(output, stmts, next_cond);
+            if (opts->loop_switch_label) {
+                UNREACHABLE();
+            } else {
+                // loop/switch mode: set `yystate` to the start DFA state of the next condition and
+                // continue to the head of the loop.
+                gen_continue_yyloop(output, stmts, next_cond);
+            }
```
skvadrik: Instead of `if/else` make it just `if` (move the main part out of `else`. Or even better, replace with a `CHECK(!opts->computed_continue)`.

src/codegen/pass2_generate.cc
```diff
@@ -1877,6 +1908,9 @@ void gen_dfa_as_switch_cases(Output& output, Adfa& dfa, CodeCases* cases) {
 
         uint32_t label = s->label->index;
         DCHECK(label != Label::NONE);
+        const char* switch_label = opts->loop_switch_label
+                ? gen_switch_label_state(output, *s->label)
+                : nullptr;
```
skvadrik: Could you move this definition under condition on line 1927 below?

src/codegen/pass2_generate.cc
```diff
@@ -1890,22 +1924,55 @@ void gen_dfa_as_switch_cases(Output& output, Adfa& dfa, CodeCases* cases) {
...
+            if (s->label->used) {
+                buf.str("    ").str(block.opts->label_prefix).label(*s->label).str(",\n");
+            }
+        }
+    }
+    buf.str("}, .").str(block.opts->label_prefix).label(start).str(")");
+
+    return buf.flush();
+}
+
```
skvadrik: This part is the most problematic.

It is Zig-specific, so it should be moved to syntax files. Here in codegen we should have only generic definitions that can be reused by multiple languages.

The best way would be to modify Zig's definition of `code:loop` to encorporate all this. It may require adding some local variables to `code:loop` configuration.

src/options/opt.cc
```diff
@@ -35,12 +35,27 @@ LOCAL_NODISCARD(Ret fix_conopt(conopt_t& glob)) {
     }
 
     if (glob.code_model != CodeModel::GOTO_LABEL) {
-        // In loop/switch or rec/func mode enable eager-skip always (not only in cases when YYFILL
-        // labels are used) to avoid special handling of initial state when there are transitions
-        // into it.
+        // In switch-based or rec/func modes enable eager-skip always (not only in cases when
+        // YYFILL labels are used) to avoid special handling of initial state when there are
+        // transitions into it.
```
skvadrik: Revert.

src/options/opt.cc
```diff
@@ -35,12 +35,27 @@ LOCAL_NODISCARD(Ret fix_conopt(conopt_t& glob)) {
...
+    if (glob.loop_switch_label) {
+        if (glob.code_model != CodeModel::LOOP_SWITCH) {
+            RET_FAIL(error("--loop-switch-label requires --loop-switch code model"));
+        }
+        if (glob.storable_state) {
+            RET_FAIL(error("storable state is not supported with --loop-switch-label"));
+        }
+        if (glob.start_conditions) {
+            RET_FAIL(error("conditions are not supported with --loop-switch-label"));
+        }
```
skvadrik, line 53: Can we rely on loop/switch model checks to avoid some of these?

src/options/opt.cc
```diff
@@ -252,6 +267,9 @@ LOCAL_NODISCARD(Ret fix_mutopt(
     if (!is_default.label_start) {
         real.label_start_force = defaults.label_start_force;
     }
+    if (glob.loop_switch_label && is_default.label_loop) {
+        real.label_loop = "state";
+    }
```
skvadrik: Why hardcode the name here? We should use some user-configurable variable like `yystate`.

src/parse/lexer.re
```diff
@@ -223,7 +223,8 @@ loop:
             RET_FAIL(error_at_cur("`getstate` without `-f --storable-state` option"));
         } else if (opts->code_model == CodeModel::LOOP_SWITCH) {
             RET_FAIL(error_at_cur(
-                    "`getstate` is incompatible with --loop-switch code model, it requires"
+                    "`getstate` is incompatible with --loop-switch code model (including"
+                    " --loop-switch-label); it requires"
```
skvadrik: Revert.

test/codegen/c/loop_switch/storable_state_error_getstate.c
```diff
@@ -1 +1 @@
-codegen/c/loop_switch/storable_state_error_getstate.re:3:16: error: `getstate` is incompatible with --loop-switch code model, it requires cross-block `goto` transitions or function calls
+codegen/c/loop_switch/storable_state_error_getstate.re:3:16: error: `getstate` is incompatible with --loop-switch code model (including --loop-switch-label); it requires cross-block `goto` transitions or function calls
```
skvadrik: Revert.

skvadrik (changes requested):
Generally, I believe there is one place that still requires a bit of work (the hardcoded Zig specifics in codegen that replace loop label should be moved to syntax file). Otherwise, it's just some renaming and polishing.

src/options/opt.cc
PongPong: No, these checks are specific to computed continue. The regular loop/switch model supports both storable state and conditions, while computed continue currently cannot represent their transitions. Therefore the existing loop/switch checks do not reject these combinations.

doc/manual/basics/syntax_files.rst_
```diff
@@ -585,6 +588,10 @@ language configurations, inplace configurations and code templates.
   ``.code_model.loop_switch``
       True if code model based on loop/switch is used (``--loop-switch``).
 
+  ``.computed_continue``
+      True if loop/switch model uses labeled-switch transitions
+      (``--computed-continue``).
+
```
skvadrik: Can you move this above line 588 to keep code models together?

src/codegen/pass2_generate.cc
```diff
@@ -478,14 +478,33 @@ void expand_fintags(Output& output, const Tag& tag, std::vector<const char*>& fi
...
+static const char* gen_yyswitch_state(Output& output, const Label& label) {
+    const opt_t* opts = output.block().opts;
+    Scratchbuf& o = output.scratchbuf;
+
+    if (opts->computed_continue) {
+        o.str(opts->label_prefix).label(label);
+    } else {
+        o.label(label);
+    }
```
skvadrik, line 502: Add a comment explaining why labels need prefixes with `--computed-continue`. Also, you could factor out the `return o.label(label).flush();` part and avoid `else` altogether.

src/codegen/pass2_generate.cc
```diff
@@ -1875,9 +1892,9 @@ void gen_dfa_as_switch_cases(Output& output, Adfa& dfa, CodeCases* cases) {
     for (State* s = dfa.head; s; s = s->next) {
         CodeList* body = code_list(alc);
 
-        uint32_t label = s->label->index;
+        const Label& state_label = *s->label;
+        uint32_t label = state_label.index;
```
skvadrik: Maybe inline this variable?

src/codegen/pass2_generate.cc
```diff
@@ -2075,7 +2103,9 @@ LOCAL_NODISCARD(Ret gen_block_code(Output& output, const Adfas& dfas, CodeList*
         // In the loop/switch mode append all DFA states as cases of the `yystate` switch.
         // Merge DFAs for different conditions together in one switch.
         local_decls = true;
-        append(code, gen_yystate_def(output, is_cond_block));
+        if (!opts->computed_continue) {
+            append(code, gen_yystate_def(output, is_cond_block));
```
skvadrik, line 2109: Add a comment explaining why `yystate` definition is omitted in the computed continue case?

src/codegen/pass4_render.cc
```diff
@@ -549,49 +549,93 @@ class RenderSwitch : public RenderCallback {
...
         switch (var) {
             case StxVarId::LABEL: rctx.os << rctx.opts->label_loop; break;
+            case StxVarId::STATE:
+                DCHECK(curr_case->kind == CodeCase::Kind::STRING);
+                rctx.os << curr_case->string;
+                break;
+            case StxVarId::VAL: rctx.os << code->init; break;
+            case StxVarId::VAR: rctx.os << rctx.opts->var_state; break;
             case StxVarId::STMT: render(rctx, curr_stmt); break;
             default: render_global_var(rctx, var); break;
```
skvadrik: Please reformat the whole switch so it has case body on a new line and dedent every case (see example of formatting on line 609 below).

src/options/opt.cc
skvadrik: Got it, thanks.

skvadrik (changes requested):

skvadrik:
Hi @PongPong , I'm sorry that I forgot to hit "submit review" a few days ago! All this time I thought I have already submitted it.

skvadrik:
It's just a few very minor things anyway.

test/config/short_option_tags.re
```diff
@@ -0,0 +1,7 @@
+// re2c $INPUT -o $OUTPUT --loop-switch --start-conditions -T
+/*!re2c
+
+<main> "a" @tag {}
+<main> *          {}
+
+*/
```
skvadrik: Drop this test? I don't think it's related to this patchset.

skvadrik (changes requested):
It looks like there are some tests failing with the latest patchset: https://github.com/skvadrik/re2c/actions/runs/32949317027/job/98119014991?pr=580 (it's probably something minor as it was passing before the latest changes).

Also, I noticed one added test that seems out of place (see the review comment).

Aside from that, all looks good and I'm happy to merge. Please squash the commits together for better git history.

PongPong (commented):
Thanks for the updates. The unrelated short_option_tags test has been removed and the commits have been squashed.

PongPong:
@skvadrik , I just ran the `linux-gcc-ubsan-ootree-full` build via docker , During configure, Automake
  generated an old-style rule combining implicit and ordinary targets. GNU Make 4.4 now rejects that deprecated
  syntax:

  *** mixed implicit and normal rules: deprecated syntax

  So the sequence was:

  1. autogen.sh succeeded.
  2. configure succeeded.
  3. The fast build started.
  4. GNU Make 4.4 stopped while parsing the generated Makefile.
  5. The four computed-continue tests were never executed.

skvadrik:
> @skvadrik , I just ran the `linux-gcc-ubsan-ootree-full` build via docker , During configure, Automake generated an old-style rule combining implicit and ordinary targets. GNU Make 4.4 now rejects that deprecated syntax:
> 
> *** mixed implicit and normal rules: deprecated syntax
> 
> So the sequence was:
> 
> 1. autogen.sh succeeded.
> 2. configure succeeded.
> 3. The fast build started.
> 4. GNU Make 4.4 stopped while parsing the generated Makefile.
> 5. The four computed-continue tests were never executed.

Why do you have to use docker? Is the problem reproducing the failures locally?

If so, you can grab some failed tests artifacts from the "Archive Test Artifacts" step in some of the failed runs. For example:

- https://github.com/skvadrik/re2c/actions/runs/32957500212/artifacts/9603081554 from https://github.com/skvadrik/re2c/actions/runs/32957500212/job/98145269592?pr=580 (CMake build)

- https://github.com/skvadrik/re2c/actions/runs/32957500256/artifacts/9603098754 from https://github.com/skvadrik/re2c/actions/runs/32957500256/job/98145271607?pr=580 (Autotools build)

Unpack the artifacts and you'll see the diffs in test results - this is often enough to fix.

On Github Actions, I don't see any Makefile-related errors (maybe the update to Make 4.4 didn't hit it yet), just the 4 failing tests.

skvadrik:
It's surprising that the tests fail in all configurations / build systems and on all platforms though.

skvadrik:
I looked at the test diffs and it seems that the 4 tests are all hitting a check at line src/codegen/output.cc:183 - the code is trying to output a label that tunrs out to be unused.

PongPong:
> I looked at the test diffs and it seems that the 4 tests are all hitting a check at line src/codegen/output.cc:183 - the code is trying to output a label that tunrs out to be unused.

i 've fixed the issue, it was caused by last refactoring after inlined the code

```
const Label& state_label = *s->label;

  because s is modified while collapsing states. Inlining *s->label later would select the final collapsed state
  —the original bug. The redundant numeric variable was inlined as state_label.index.
 ```

skvadrik (approved):

skvadrik:
All good! Can you squash commits before merging?

skvadrik:
> > I looked at the test diffs and it seems that the 4 tests are all hitting a check at line src/codegen/output.cc:183 - the code is trying to output a label that tunrs out to be unused.
> 
> i 've fixed the issue, it was caused by last refactoring after inlined the code
> 
> ```
> const Label& state_label = *s->label;
> 
>   because s is modified while collapsing states. Inlining *s->label later would select the final collapsed state
>   —the original bug. The redundant numeric variable was inlined as state_label.index.
> ```

Ah, I wondered about that. Thanks for the explanation.

PongPong (commented):
i've squashed into single commit.

skvadrik:
Perfect. I'm really happy with the final implementation. Thanks for all your work and patience addressing the comments!

PongPong:
Hi @skvadrik , do you have any plans to release a new stable version of re2c soon?

skvadrik:
> Hi @skvadrik , do you have any plans to release a new stable version of re2c soon?

Sure, I can release 4.6 with the new feature.

skvadrik:
https://github.com/skvadrik/re2c/releases/tag/4.6 (release notes a bit late but I'm on it).

Issue #545, opened 2025-05-13
Broadword parsing


terpstra:
Would it be feasible to enhance re2c to process more than a single character a time? And/or optimize the parser to use less conditional branches?

Board-word techniques allow you to process multiple characters at once. An example of the sort of thing I mean:
- https://arxiv.org/abs/1902.08318
- or, some examples I wrote:

```
// A decent compiler will turn this into a single 8B load
static uint64_t grab8(const unsigned char *arg) {
  return static_cast<uint64_t>(arg[0]) | static_cast<uint64_t>(arg[1]) << 8 |
         static_cast<uint64_t>(arg[2]) << 16 | static_cast<uint64_t>(arg[3]) << 24 |
         static_cast<uint64_t>(arg[4]) << 32 | static_cast<uint64_t>(arg[5]) << 40 |
         static_cast<uint64_t>(arg[6]) << 48 | static_cast<uint64_t>(arg[7]) << 56;
}

static uint32_t hex2uint(const char *arg) {
  uint64_t x = grab8(reinterpret_cast<const unsigned char *>(arg));
  x = (x & 0x0f0f0f0f0f0f0f0fULL) + 9 * ((x & 0x4040404040404040ULL) >> 6);
  x = ((x << 4) | (x >> 8)) & 0x00ff00ff00ff00ffULL;
  x = ((x << 8) | (x >> 16)) & 0x0000ffff0000ffffULL;
  x = ((x << 16) | (x >> 32));
  return x;
}

static bool legalhex(const char *arg) {
  uint64_t lo = 0x0101010101010101ULL;
  uint64_t hi = 0x8080808080808080ULL;
  uint64_t y = grab8(reinterpret_cast<const unsigned char *>(arg));
  uint64_t x = y & ~hi;
  uint64_t z = ~(x + lo * (128 - '0')) | (x + lo * (127 - '9'));
  uint64_t a = ~(x + lo * (128 - 'a')) | (x + lo * (127 - 'f'));
  uint64_t A = ~(x + lo * (128 - 'A')) | (x + lo * (127 - 'F'));
  return !(((z & a & A) | y) & hi);
}
```

To do this well, one probably needs some way to teach re2c which the most common / expected paths through the parser state will be encountered in the real world. That doesn't seem too difficult, though; one could supply re2c with real world training files to guide its choices.

terpstra:
For example, consider this code extracted from the inner loop of a JSON5 string parser:

```
    /*!re2c
[...]

PR #585, opened 2026-09-12
Go: SIMD fast-forward loop for reflexive DFA base states (#411)

### LINKED ISSUE #411: Is there SIMD support?
For example, use SIMD intrinsics explicity, or use long long to process 8 bytes together?

hazyhaar:
### Summary
Resolves #411 by introducing SIMD hardware vectorization for absorbing reflexive loops in regular expressions, targeting the Go backend via Go 1.27's standard `simd/archsimd` package.

### Motivation & Background
In high-throughput lexical analyzers (e.g. string literals, identifiers, long whitespace sequences), the lexer spends most CPU cycles in tight loops over homogenous byte classes. Conventional DFAs execute this byte-by-byte using conditional branches.

### Architecture & Design
1. **Reflexive State Analysis (`Adfa::find_simd_classes`)**:
   - Detects absorbing reflexive loops where all transitions within a character class loop back to the same state.
   - Strictly respects the `MATCH`/`MOVE` split (`move->prev == match`) and tag hoisting invariants (`s->go.tags == TCID0`, `to->save == NOSAVE`).
   - If tags or backtrack recording are present on the loop, vectorization is safely bypassed.

2. **Go 1.27 Native Vectorization (`simd/archsimd`)**:
   - Emits a 32-byte fast-forward preamble using `archsimd.LoadUint8x32`, vector comparisons, and `bits.TrailingZeros32` for exact byte-accurate delimiter positioning.
   - Zero heap allocations (**0 B/op**) guaranteed.
   - Clean fallback to scalar DFA for remaining bytes (< 32 bytes) and unaligned buffers.

3. **Opt-in Control**:
   - Enabled via `--simd` CLI flag or `re2c:simd = 1;` in-source configuration.

### Verification & Metrology
- **1,565 / 1,565 test cases passed** with zero regressions (`run_tests.py`).
- **Parity Oracle**: Validated bit-exact on 6,405 inputs against scalar reference (`examples/go/simd/simd.re`).
- **Real-World Example**: Added realistic escaped string scanner in `examples/go/simd/01_string_lexer.re` with zero-allocation test harness (`01_string_lexer_test.go`).
- **Documentation**: Added Sphinx manual chapter (`doc/manual/features/simd/simd.rst_`) and manpage entry.


skvadrik:
Hi! First, thanks a lot for the contribution. I'm not keeping silent because I have't seen it, but because I need more time to think it over. It's been on my mind lately as well, so I'm glad to see it materialized in the form of a PR.

This is not a detailed code review yet - I'd like to discuss generic approach first.

Regarding option and configuration names: #585 and #586 are related, so I'd like to have a common family of options like `--vectorize-loops`, `--vectorize-linear` and so on rather than completely unrelated option names. Then perhaps in the future we could have `--vectorize-all` that would enable all of them.

The main blocker is the need for generalization over all language backends. Things like `gen_simd` with hardcoded language-specific code are unacceptable. The current approach in re2c is to push all language-specific stuff out of codegen and into syntax files. So, we'll need new syntax configuration(s) that will be generic enough to allow definition in multiple languages. Even if this PR only handles Go, we should have the broad idea / draft how it can be implemented for other languages. It will take some time to figure out the right generalization.

Finally, for such a performance-related PR I'd like to see some benchmarks that prove it's making things faster.

Let's focus the discussion here for both #586 and #585 (this one).

hazyhaar:
Hi @skvadrik,

Thank you very much for the insightful feedback. We have reworked both PRs (#585 and #586) to address all points raised:

### 1. Unified Option & Configuration Nomenclature
We have aligned both features under a cohesive naming family:
- **Loop vectorization (#411 / #585):** Renamed `--simd` to `--vectorize-loops`, and configuration `re2c:simd` to `re2c:vectorize:loops` (with `re2c:flags:vectorize:loops`).
- **Linear path coalescing (#545 / #586):** Renamed `--multi-char` to `--vectorize-linear`, and configuration `re2c:flags:multibyte` to `re2c:vectorize:linear`.
- Both flags are documented under this common nomenclature in the Sphinx manual and manpages, paving the way for an umbrella `--vectorize-all` option once merged.

### 2. Multi-Backend Generalization (Zero Language-Specific Hardcoding in Codegen)
We have completely extracted the hardcoded Go code out of `src/codegen/pass2_generate.cc`:
- **AST Representation:** Introduced `CodeKind::VECTOR_LOOP` and a language-agnostic `CodeVectorRanges` container in `src/codegen/code.h` storing the reflexive class intervals $[lo_i, hi_i]$ (supporting up to 8 disjoint character class ranges).
- **Generic Rendering Callback:** In `src/codegen/pass4_render.cc`, `RenderVectorLoop` drives the emission through a new template configuration `code:vector_loop` in `include/syntax/`.
- **Target Implementations:**
  - **Go 1.27:** Defined in `include/syntax/go` using the standard `simd/archsimd` package (`LoadUint8x32`, `GreaterEqual`/`LessEqual`, `Or`, and bitmask check).
  - **C:** Defined in `include/syntax/c` using standard portable AVX2 intrinsics (`_mm256_loadu_si256`, unsigned saturated subtraction `_mm256_subs_epu8` with `_mm256_cmpeq_epi8`, and `_mm256_movemask_epi8`).
  - **Other 10 backends:** Explicitly defined as `<undefined>` with `re2c:vectorize:loops = 0` until syntax templates are provided.
- Regenerated bootstrap headers across all 12 backends.

### 3. Buffer Boundary & Safety Hardening
Following an in-depth audit, two critical memory safety invariants have been resolved and verified with AddressSanitizer:
1. **Scalar Reserve Margin in Vector Loops:** Guard condition now enforces $\text{available} \ge \text{VECTOR\_SIZE} + \max(1, s\to\text{fill})$, guaranteeing that after advancing by 32 bytes, at least 1 valid byte remains in the buffer for the subsequent scalar peek without exceeding buffer boundaries.
2. **Broadword Stack & Read Safety:** Fixed an off-by-one stack array allocation in `Adfa::coalesce_multichar` (`path[9]` for 8 transitions), and secured `YYPEEKN` against over-reading past null sentinels on short inputs.

### 4. Empirical Benchmark Measurements (Intel Core i9-14900K, Linux amd64)
All benchmarks were performed with real data and verified bit-exact against scalar oracles with 0 memory allocation:

#### A. Reflexive Loops (`--vectorize-loops` / #411)
- **Workload:** String tokenization with escaped delimiters (`examples/go/simd/01_string_lexer_test.go`).
- **Scalar baseline:** 1,610.04 MB/s (1.61 GB/s) — 0 allocs/op
- **Vectorized loop:** **5,777.22 MB/s (5.78 GB/s)** — 0 allocs/op
- **Speedup:** **3.59x** (+258.8% throughput, -72.1% execution time).
- **Bit-exact parity:** 6,405 test inputs verified in `test/codegen/go/004_simd.re` and C AVX2 test in `test/codegen/c/vectorize_loops.re`.

#### B. Linear Broadword (`--vectorize-linear` / #545)
- **Workload:** SQL keyword recognition (`examples/go/multibyte/01_keywords_test.go`).
- **Scalar baseline:** 2.44 ns/op (2,866 MB/s)
- **Direct 32-bit machine word:** **0.14 ns/op (49,347 MB/s)**
- **Speedup:** **17.2x** at 0 allocs/op.

Both PR branches have been updated and are ready for your review.

